PR #21, opened 2015-01-14
chore(l10n): setup extraction script for l10n

### LINKED ISSUE #20: export the templates to l10n
move the export of the email templates from content-server to here

zaach:
Fixes #20.

GRUNT :bomb: 

@dannycoates r?

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -1,0 +1,59 @@
+#!/usr/bin/env node
+
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+const fs = require('fs');
+const path = require('path');
+const extract = require('jsxgettext-recursive');
+
+// # Globbing
+// for performance reasons we're only matching one level down:
+// 'test/spec/{,*/}*.js'
+// use this if you want to recursively match all subfolders:
+// 'test/spec/**/*.js'
+module.exports = function (grunt) {
+  'use strict';
+
+  grunt.loadNpmTasks('grunt-contrib-copy');
+
+  grunt.initConfig({
+    copy: {
+      strings: {
+        files: [{
+          expand: true,
+          flatten: true,
+          cwd: path.join(__dirname, 'node_modules', 'fxa-content-server-l10n', 'locale', 'templates', 'LC_MESSAGES'),
+          dest: __dirname,
+          src: [
+            'server.pot'
+          ]
+        }]
+      }
+    }
+  });
+
+  grunt.registerTask('l10n-extract', 'Extract strings from templates for localization.', function () {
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -20,7 +20,7 @@
   "dependencies": {
     "bluebird": "2.2.2",
     "bunyan": "1.0.0",
-    "fxa-content-server-l10n": "git://github.com/mozilla/fxa-content-server-l10n.git#4168ef3cb7",
+    "fxa-content-server-l10n": "git://github.com/mozilla/fxa-content-server-l10n.git",
```
zaach: Hrm. We can keep this locked down but it needs to be updated to the latest hash before extracting strings. This is so the extracted strings are merged with the latest template from the l10n repo.

@dannycoates What's the significance of the hash this is currently locked to? Is there a policy for updating it?

Gruntfile.js
```diff
@@ -0,0 +1,62 @@
...
+
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+// Generated on 2013-12-04 using generator-backbone-amd 0.0.4
+//
+const fs = require('fs');
+const path = require('path');
+const mkdirp = require('mkdirp');
```
pdehaan: Is `mkdirp` used somewhere and I'm not seeing it?

Gruntfile.js
```diff
@@ -0,0 +1,62 @@
...
+  grunt.loadNpmTasks('grunt-contrib-copy');
+
+  grunt.initConfig({
+    copy: {
+      strings: {
+        files: [{
+          expand: true,
+          flatten: true,
+          cwd: path.join(__dirname, 'node_modules', 'fxa-content-server-l10n', 'locale', 'templates', 'LC_MESSAGES'),
+          dest: path.join(__dirname),
```
pdehaan: Not sure if we need the `path.join()` dance if there is only one directory (same w/ line 45 below)

Gruntfile.js
```diff
@@ -0,0 +1,59 @@
+#!/usr/bin/env node
+
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+const fs = require('fs');
```
pdehaan: Not sure if we're using `fs` anywhere.

package.json
dannycoates: I think we might be able to set this free. It was locked down before because this repo didn't "control" its own strings so we locked it to a known good commit. Now, as long as we follow the workflow in the readme we should be ok. *As long as we don't run npm install on and outdated version of this repo* (but that's not how we roll... literally)

dannycoates:
:+1: solid r+

PR #32, opened 2015-03-11
fix(l10n): Ensure the email subjects are extracted.

### LINKED ISSUE #31: [train-32] dropped email subjects 'Verify your account' and 'Reset your password'
Github doesn't handle the big string change diffs, so look at, for example, `git show 3d39360 locale/fr/LC_MESSAGES/server.po` in fxa-content-server-l10n

That picked up some strings, noticeably "Almost there" that were missing, but appears to have dropped the email subjects.

shane-tomlinson:
Look in the .js files for strings surrounded by `gettext()`

fixes #31

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -50,7 +50,25 @@ module.exports = function (grunt) {
     });
 
     walker.on('end', function () {
-      done();
+      var jsWalker = extract({
+        'input-dir': __dirname,
+        /* node_modules and test should not contain any strings
+         * Gruntfile causes an error and should contain no strings
+         * bin/server.js extracts "/", so it is excluded.
+         */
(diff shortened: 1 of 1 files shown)


Gruntfile.js
```diff
@@ -50,7 +50,25 @@ module.exports = function (grunt) {
     });
 
     walker.on('end', function () {
-      done();
+      var jsWalker = extract({
+        'input-dir': __dirname,
+        /* node_modules and test should not contain any strings
+         * Gruntfile causes an error and should contain no strings
+         * bin/server.js extracts "/", so it is excluded.
+         */
```
rfk: *grumble* if the `.js` files were all in a `./lib` directory we woud not have this problem...

rfk:
r+

PR #38, opened 2015-04-14
add auth obj in options if set in config

sovanna:
in local development, I've updated `/fxa-auth-server/config/dev.json`. adding into smtp, `user` and `password` as they are by default `undefined` (trying to use smtp from Gandi.net)

but I've received `RecipientError` `Client host rejected: Access denied` by the server

Then I saw that these prop in config are not used in nodemailer when it is called.. that's why I've added the `auth` config.

### DIFF
--- a/mailer.js
+++ b/mailer.js
@@ -14,6 +14,14 @@ module.exports = function (log) {
       secureConnection: config.secure,
       port: config.port
     }
+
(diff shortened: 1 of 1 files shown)


rfk:
LGTM, thanks @sovanna!

Issue #39, opened 2015-04-22
[train-35][prod] 500 errors due to change in defaultLanguage


jrgm:
With train-35 going to prod, there were ~60 500 errors in a 20 minute window before DNS was switched back to train-34 for fxa-auth-server. It was a little confusing at first seemed all errors were from Android. 

```
TypeError: Cannot call method 'gettext' of undefined 
at Mailer.verifyEmail (/data/fxa-auth-server/node_modules/fxa-auth-mailer/mailer.js:73:27) 
at Mailer.mailer.sendVerifyCode (/data/fxa-auth-server/mailer.js:21:25) 
at customs.check.then.then.then.then.then.done.account (/data/fxa-auth-server/routes/account.js:164:24) 
at HandleCallback (/data/fxa-auth-server/node_modules/p-promise/p.js:506:8) 
at Then (/data/fxa-auth-server/node_modules/p-promise/p.js:493:5) 
at Object.flush [as _onImmediate] (/data/fxa-auth-server/node_modules/p-promise/p.js:204:4) 
at processImmediate [as _immediateCallback] (timers.js:354:15)
```

At any rate, this boils down to this line:
https://github.com/mozilla/fxa-auth-mailer/blob/master/translator.js#L43
where a default of `en-US` is listed, while the fxa-content-server-l10n switched from a default of `en-US` to `en` in this release. Simply changing the line above to say `en` fixes this problem (of course it would be better if we injected the defaultLanguage from fxa-auth-server into this module, and put in a regression test).

Simple STR, is to configure your browser to use some language we don't support, e.g., `haw` for Hawaiian, and try to create an account.

jrgm:
Draft of a fix at https://github.com/mozilla/fxa-auth-mailer/compare/master...jrgm:fix-l10n-aargh

shane-tomlinson:
@jrgm already has a PR for this.

jrgm:
So fixed in https://github.com/mozilla/fxa-auth-mailer/pull/40

PR #40, opened 2015-04-22
fix(l10n): make defaultLanguage configurable, and switch to "en" default

jrgm:
This adjusts the defaultLanguage from 'en-US' to 'en' and makes that sorta configurable. 
I don't particularly like the `config.i18n` part, since that's a detail from fxa-auth-server.

### DIFF
--- a/config.js
+++ b/config.js
@@ -3,7 +3,8 @@ module.exports = require('rc')(
   {
     port: 1810,
     logLevel: 'info',
-    locales: ['en_US', 'de'],
+    locales: ['en', 'de'],
+    defaultLanguage: 'en'
     mail: {
       host: '127.0.0.1',
       port: 9999,
--- a/index.js
+++ b/index.js
@@ -9,7 +9,7 @@ module.exports = function (log, config) {
   var Mailer = createMailer(log)
   return P.all(
     [
-      require('./translator')(config.locales),
+      require('./translator')(config.locales, config.defaultLanguage),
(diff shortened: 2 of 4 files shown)


jrgm:
Okay, took me a couple of takes, but this is simpler. 

Feel free to point out how wrong I am. r? dannycoates

jrgm:
fxa-auth-server will need to add defaultLanguage here https://github.com/mozilla/fxa-auth-server/blob/master/mailer.js#L14

jrgm:
cross ref https://github.com/mozilla/fxa-auth-mailer/issues/39

rfk:
Thanks @jrgm.  I wonder if we could simplify this by assuming that `locales[0]` is the default language.  Maybe too much magic though.

Also /cc @zaach for reference, since I mentioned this in our 1:1 today.

rfk:
> fxa-auth-server will need to add defaultLanguage

Like so?  https://github.com/mozilla/fxa-auth-server/pull/917

This LGTM, r+.

jrgm:
> Like so? mozilla/fxa-auth-server#917

Yeah, that's what I meant. I was also going to add a test to call the translator with all the languages (and others).

rfk:
Please, add as many tests as you like :-)

dannycoates:
@jrgm yep LGTM r+

jrgm:
Thanks.

Issue #36, opened 2015-04-13
support schema.org "confirm action" for account verification


zaach:
Inspired by Ryan's link in #35, we could help GMail users out by supporting the schema.org confirm action. This would allow users to verify the account in one click from their GMail inbox.
https://developers.google.com/gmail/markup/reference/one-click-action#confirm_action

shane-tomlinson:
This is an awesome idea.

jrgm:
So Shane, do you want to take this?

zaach:
We'd also have to register with Google: https://developers.google.com/gmail/markup/registering-with-google

pdehaan:
Not sure if this is related or useful, but I thought this schema thing sounded familiar: https://github.com/mozilla/persona/pull/4057.

rfk:
This might fit nicely under @eoger's general remit of "improve verification rate".

rfk:
Great stuff here folks!   IIUC this will not be active until we do some registration with Google for our production servers, right?

@eoger, what metrics should we watch out for when this goes live?  I think it would make an interesting blog post or similar if we can say "increased verification by X%".

vladikoff:
```javascript
query.one_click = true
var oneClickLink = this.verificationUrl + '?' + qs.stringify(query)
```
@rfk the one click link has an extra `one_click=true` parameter. I have no clue how we can extract the data and compare it with other clicks.

rfk:
It would be neat to run this as an a/b test but probably too much mucking around to make it worthwhile.  I'm hoping it will show up as a measurable improvement in our verification rate, but we might need to try to limit it to "improved verification rate for gmail addresses" to make it visible.  I dunno, just thinking out loud mostly :-)

PR #42, opened 2015-06-05
feat(signup): add schema.org one-click-action button to verify email

eoger:
Fixes https://github.com/mozilla/fxa-auth-mailer/issues/36.

screen shot 2015-06-04 at 5 10 44 pm
Right now to test it you have to copy-paste the email HTML code [here](http://gmail-actions.appspot.com/).

I think maybe @jrgm will have to [register to Google](https://developers.google.com/gmail/markup/registering-with-google) to make it work on production servers.

### DIFF
--- a/mailer.js
+++ b/mailer.js
@@ -69,10 +69,13 @@ module.exports = function (log) {
     if (message.resume) { query.resume = message.resume }
 
     var link = this.verificationUrl + '?' + qs.stringify(query)
+    query.one_click = true
+    var oneClickLink = this.verificationUrl + '?' + qs.stringify(query)
 
     var values = {
       translator: translator,
       link: link,
+      oneClickLink: oneClickLink,
       email: message.email
     }
     var localized = this.templates.verifyEmail(values)
--- a/templates/verify.html
+++ b/templates/verify.html
@@ -31,6 +31,14 @@
     </style>
   </head>
   <body leftmargin="0" marginwidth="0" topmargin="0" marginheight="0" offset="0" style="background-color:#F2F2F2;width:100% !important;height:100% !important;padding:0;margin:0;-ms-text-size-adjust:100%;-webkit-text-size-adjust:100%;">
+    <div itemscope itemtype="https://schema.org/EmailMessage">
+      <div itemprop="potentialAction" itemscope itemtype="https://schema.org/ViewAction">
+        <link itemprop="target" href="{{{oneClickLink}}}"/>
+        <meta itemprop="name" content="{{t 'Verify Email'}}"/>
+        <meta itemprop="url" content="{{{oneClickLink}}}"/>
+      </div>
+      <meta itemprop="description" content="{{t 'Verify your email to finish your Firefox Account registration'}}" />
+    </div>
     <center>
       <table align="center" border="0" cellpadding="0" cellspacing="0" height="100%" width="100%" id="bodyTable" style="background-color:#F2F2F2;width:100% !important;padding:0;margin:0;height:100% !important;border-collapse:collapse !important;mso-table-rspace:0pt;mso-table-lspace:0pt;-ms-text-size-adjust:100%;-webkit-text-size-adjust:100%;">
         <tr>


templates/verify.html
```diff
@@ -31,6 +31,14 @@
     </style>
   </head>
   <body leftmargin="0" marginwidth="0" topmargin="0" marginheight="0" offset="0" style="background-color:#F2F2F2;width:100% !important;height:100% !important;padding:0;margin:0;-ms-text-size-adjust:100%;-webkit-text-size-adjust:100%;">
+    <div itemscope itemtype="http://schema.org/EmailMessage">
+      <div itemprop="potentialAction" itemscope itemtype="http://schema.org/ViewAction">
```
pdehaan: I don't know nothing about HTML emails, but do we want these to be HTTP or HTTPS? Both seem to resolve correctly on schema.org. Wasn't sure if there would be potential issues w/ an HTTPS mail client, or probably not since I'm guessing these aren't downloading specific resources.

templates/verify.html
```diff
@@ -31,6 +31,14 @@
...
+        <link itemprop="target" href="{{{link}}}"/>
+        <meta itemprop="name" content="{{t 'Complete sign-up'}}"/>
```
vladikoff: Seems like we are using double quotes in other strings

vladikoff:
ping @ryanfeeley , what wording would you like to see here? 

"Complete sign-up" , "Verify Email", other options?

templates/verify.html
pdehaan: Wouldn't double quotes break the nesting here? (ie: close the `content` attribute)

templates/verify.html
```diff
@@ -31,6 +31,14 @@
...
+        <meta itemprop="url" content="{{{link}}}"/>
```
vladikoff: is there a way to see that people are actually using this gmail button instead of the regular link? thoughts?

templates/verify.html
vladikoff: That's true, I personally don't mind, just pointing it out. @zaach thoughts?

templates/verify.html
eoger: URL param maybe?

vladikoff:
ping @ryanfeeley , see https://github.com/mozilla/fxa-auth-mailer/pull/42#issuecomment-109102122

templates/verify.html
shane-tomlinson: > URL param maybe?

Hey, that's what I just suggested in the IRC channel.

eoger: So what should we do here? Modify the URL in this PR and open another issue for the actual processing of the new URL param ?

ryanfeeley:
"Verify Email" because that's the step being completed for sure. Maybe in future sign-up will require more.

templates/verify.html
vladikoff: If no one has any other ideas, let's just add `one_click=true` to that url. I guess we can track it in the server logs for now. just make sure the generated url with `one_click` is a valid url.

eoger:
Updated!

eoger:
The Travis build passed, what do we do with this PR?

vladikoff:
> I think maybe @jrgm will have to register to Google to make it work on production servers.

@eoger can you confirm that this can be safely merged without registering ^

If so then we should file a bug on Bugzilla and notify the ops team to configure that for us. @ckolos might know howe we can proceed with this.

eoger:
I'm 99% sure it will not break anything on merge but I can't find anywhere on their documentation saying the emails will not be rejected if we're not registered yet.
On the practical side, I tried it the patch locally and I received the e-mail anyway (but the schema.org markups were not processed though).
Should I contact someone from Gmail just to be sure?

templates/verify.html
vladikoff: @eoger could we update this to use https to avoid being rejected as an email with http:// links ?

mailer.js
```diff
@@ -68,10 +68,13 @@ module.exports = function (log) {
     if (message.resume) { query.resume = message.resume }
 
     var link = this.verificationUrl + '?' + qs.stringify(query)
+    query.one_click = "true"
```
vladikoff: can we safely make this a true `true` instead of a string?
it gets `qs.stringify(query)`'d anyway

templates/verify.html
```diff
@@ -31,6 +31,14 @@
     </style>
   </head>
   <body leftmargin="0" marginwidth="0" topmargin="0" marginheight="0" offset="0" style="background-color:#F2F2F2;width:100% !important;height:100% !important;padding:0;margin:0;-ms-text-size-adjust:100%;-webkit-text-size-adjust:100%;">
+    <div itemscope itemtype="http://schema.org/EmailMessage">
+      <div itemprop="potentialAction" itemscope itemtype="http://schema.org/ViewAction">
+        <link itemprop="target" href="{{{oneClickLink}}}"/>
+        <meta itemprop="name" content="{{t 'Verify Email'}}"/>
```
vladikoff: :+1:

vladikoff:
:+1:

vladikoff:
@eoger could you file a bugzilla bug so we can discuss setting this up for prod?  In BugZilla: cc me, @jrgm and @ckolos and mark it "Confidential" (it's an option under "Advanced Fields"). 

Let me know if you have any questions.

eoger:
On it, thanks for the merge!

shane-tomlinson:
Has a bugzilla bug been filed to set this up for prod? cc @rfk, @eoger, @vladikoff and @philbooth

vladikoff:
> Has a bugzilla bug been filed to set this up for prod? cc @rfk, @eoger, @vladikoff and @philbooth

Yeah there is a bugzilla bug but I cannot find it @jrgm might have it listed

eoger:
Here you go: [Bugzilla bug #1176102](https://bugzilla.mozilla.org/show_bug.cgi?id=1176102)

Issue #45, opened 2015-06-11
Add ESLint to repo (as a grunt task)


pdehaan:
Testing out my new [**eslint-config-fxa**](https://github.com/pdehaan/eslint-config-fxa/) shareable ESLint config, and noticed a few errors and warnings in this repo:

### Steps to repro:
1. `$ npm i eslint eslint-config-fxa -D`
2. Create an /.eslintrc file with the following magic YAML sauce:
    ```yaml
extends: fxa/server

rules:
      strict: 0
```
2. `$ eslint .`

### Actual results:
```
$ eslint .

Gruntfile.js
   7:6   warning  fs is defined but never used  no-unused-vars
   7:25  warning  Extra semicolon               semi
   8:29  warning  Extra semicolon               semi
   9:48  warning  Extra semicolon               semi
  17:15  warning  Extra semicolon               semi
  19:43  warning  Extra semicolon               semi
  35:5   warning  Extra semicolon               semi
  38:28  warning  Extra semicolon               semi
  50:7   warning  Extra semicolon               semi
  67:9   warning  Extra semicolon               semi
  70:15  warning  Extra semicolon               semi
  71:9   warning  Extra semicolon               semi
  72:7   warning  Extra semicolon               semi
  73:5   warning  Extra semicolon               semi
  75:68  warning  Extra semicolon               semi
  77:2   warning  Extra semicolon               semi

bin/server.js
  45:72  error  "body" is not defined                             no-undef
  71:4   error  Don't use process.exit(); throw an error instead  no-process-exit

mailer.js
  24:4  error  Trailing spaces not allowed  no-trailing-spaces

templates/index.js
  4:35  error  Strings must use singlequote  quotes

test/local/translator_tests.js
  3:33  error  A space is required after ','  comma-spacing

test/mail_helper.js
  24:4   warning  mail is defined but never used  no-unused-vars
  33:16  error    Strings must use singlequote    quotes
  39:12  warning  uid is defined but never used   no-unused-vars

translator.js
  35:64  error  Strings must use singlequote  quotes

✖ 25 problems (7 errors, 18 warnings)
```

### Expected results:
Less errors/warnings. Most are nonsense and single quotes or semi-colon consistency, but the /bin/server.js `body` one looks like a legit bug.

**Comment by pdehaan** [comment]
I fixed most of these w/ #46 PR. New output should be:
```
$ eslint .

bin/server.js
  45:72  error  "body" is not defined  no-undef

[...]

PR #46, opened 2015-06-11
chore(cleanup): Fixed some syntax errors reported by ESLint

pdehaan:
Addresses some of the basic ESLint style issues (quotes, spaces, semicolons) reported in #45.

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -4,9 +4,8 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
-const fs = require('fs');
-const path = require('path');
-const extract = require('jsxgettext-recursive');
+const path = require('path')
+const extract = require('jsxgettext-recursive')
 
 // # Globbing
 // for performance reasons we're only matching one level down:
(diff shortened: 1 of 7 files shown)


pdehaan:
Not sure if I broke the tests. I doubt it because I mostly added a space and changed quotes in the related files. Plus I notice that #42 PR was failing too with the same errors.

jrgm:
Yeah, I'd actually picked this up in other tests this afternoon (scripts/e2e-email and fxa-auth-server tests). 
The reason is that the fxa-content-server-l10n localization for pt_BR changed their `Language:` header from `pt_BR` to `pt-BR`, which is counter to all the other country-specific localizations in fxa-content-server-l10n (es_AR,es_CL,nb_NO,sr_LATN,sv_SE,zh_CN,zh_TW) and the code in this repo is not prepared to deal with that (but shouldn't crash, of course). 

I'll file a bug in fxa-content-server-l10n asking for this change to be reversed if the change was incorrect. 
I'll look at a patch for this repo.

jrgm:
The issue with crashing due to l10n content has been fixed, so ignore my earlier comment.

pdehaan:
Closing.
This needs to be rebased, and @TDA did a better job with his #50 PR.

PR #47, opened 2015-06-13
fix(locales): always normalize the locale

jrgm:
So this fixes the issue noted in https://github.com/mozilla/fxa-auth-mailer/pull/46#issuecomment-111657709 and https://github.com/mozilla/fxa-content-server-l10n/issues/52

(This whole 'upper-case or lower-case', '- or _', '2 or 3 or 4 character' language and/or country-code is fugly).


dannycoates:
r+

PR #48, opened 2015-06-15
style(verify): add logo and restyle verify email

johngruen:
Finally got this thing humming!

Here's a bunch of screenshots of how it looks everywhere: 
https://www.dropbox.com/sh/08l1mgm583tc746/AACJyIfyyvwkAygDMKtsR0rza?dl=0

@ryanfeeley

### DIFF
--- a/templates/verify.html
+++ b/templates/verify.html
@@ -1,118 +1,64 @@
 <!DOCTYPE html PUBLIC "-//W3C//DTD XHTML 1.0 Transitional//EN" "http://www.w3.org/TR/xhtml1/DTD/xhtml1-transitional.dtd">
-<html>
+<html xmlns="http://www.w3.org/1999/xhtml">
   <head>
-    <meta http-equiv="Content-Type" content="text/html; charset=UTF-8">
-    <title>{{t "Firefox Accounts"}}</title>
-    <style>
-      a {
-        color: #0095dd;
-        text-decoration: none !important;
-      }
-
-      @media only screen and (max-width: 599px){
-        /*CLIENT-SPECIFIC MOBILE STYLES */
-        body, table, td, p, a, li, blockquote{-webkit-text-size-adjust:none !important;} /* Prevent Webkit platforms from changing default text sizes */
-        body{width:100% !important; min-width:100% !important;} /* Prevent iOS Mail from adding padding to the body */
-        #bodyCell{padding:10px !important;}
-
-        .flexibleContainer {
-          width:85% !important;
-        }
-
-        #templateContainer{
-          max-width:600px !important;
-          width:100% !important;
-        }
-        h1{
-          font-size:24px !important;
-          line-height:100% !important;
-        }
-      }
-    </style>
+    <meta http-equiv="Content-Type" content="text/html; charset=UTF-8" />
+    <title>{{t "Firefox Account Activation"}}</title>
   </head>
-  <body leftmargin="0" marginwidth="0" topmargin="0" marginheight="0" offset="0" style="background-color:#F2F2F2;width:100% !important;height:100% !important;padding:0;margin:0;-ms-text-size-adjust:100%;-webkit-text-size-adjust:100%;">
-    <div itemscope itemtype="https://schema.org/EmailMessage">
-      <div itemprop="potentialAction" itemscope itemtype="https://schema.org/ViewAction">
-        <link itemprop="target" href="{{{oneClickLink}}}"/>
-        <meta itemprop="name" content="{{t 'Verify Email'}}"/>
-        <meta itemprop="url" content="{{{oneClickLink}}}"/>
-      </div>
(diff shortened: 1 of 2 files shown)


johngruen:
@ryanfeeley could you take a look at the screenshots linked above and also confirm the language in https://github.com/johngruen/fxa-auth-mailer/blob/issue-35/templates/verify.txt

ryanfeeley:
Looks great! Can't wait to get the numbers.

rfk:
Looks good!  I guess this overwrites some (but not all) of the changes in https://github.com/mozilla/fxa-auth-mailer/pull/55/files.  Should we consider applying the change from that bug ("verify" => "verify now") to the new button here as well?

(and just to note explicitly, this has new strings so let's wait until after trian-41 to merge it)

rfk:
> Can't wait to get the numbers.

Do we have plans to full-on A/B test this thing?

johngruen:
@vladikoff not sure if/how we are recording metrics from emails except to look at verification percentage

templates/verify.html
```diff
@@ -1,118 +1,64 @@
 <!DOCTYPE html PUBLIC "-//W3C//DTD XHTML 1.0 Transitional//EN" "http://www.w3.org/TR/xhtml1/DTD/xhtml1-transitional.dtd">
-<html>
+<html xmlns="http://www.w3.org/1999/xhtml">
   <head>
     <meta http-equiv="Content-Type" content="text/html; charset=UTF-8">
```
pdehaan: it this valid xhtml, or do we need to close this meta using `<meta />`?

johngruen: @pdehaan u r right dawg

templates/verify.html
```diff
@@ -1,118 +1,64 @@
...
+  <body style="-ms-text-size-adjust: 100%; -webkit-text-size-adjust: 100%; margin: 0; padding: 0;">
+    <table border="0" cellpadding="0" cellspacing="0" height="100%" width="100%" id="body-table" style="-webkit-text-size-adjust: 100%; border-collapse: collapse; mso-table-lspace: 0pt; mso-table-rspace: 0pt;">
+      <tr>
+        <td align="center" valign="top">
+          <table align="center" border="0" cellpadding="0" cellspacing="0" width="310" id="email-container" style="-webkit-text-size-adjust: 100%; border-collapse: collapse; mso-table-lspace: 0pt; mso-table-rspace: 0pt; width: 310px; margin: 0 auto;">
+            
+            <!--Logo-->
+            <tr>
+              <td align="center" id="firefox-logo" style="padding: 20px 0;">
+                <img src="http://image.e.mozilla.org/lib/fe9915707361037e75/m/2/fxlogojg.gif" height="95" width="88" alt="Firefox" style="-ms-interpolation-mode: bicubic;">
```
pdehaan: Same as above, since we're saying XHTML 1.0, does this need to be ``?
I'll see if I can run this through some HTML linter.

johngruen:
@pdehaan cleaned up those XHTML empty tags. The only validation issues i now see are with non-standard attributes (height, border etc) which are used in nearly all email templates because of non-standard rendering

templates/verify.txt
```diff
@@ -1,6 +1,6 @@
-{{t "Almost there"}}
+{{t "Welcome!"}}
```
rfk: @ryanfeeley did we discover that exclamation marks tend to be a bit of a spam trigger?  I have vague recollections of removing them long ago, but it might have had more to do with the whole phrase "Congratulations!" than with just exclamation marks in general.

rfk:
@johngruen what do you think about apply the advice from https://github.com/mozilla/fxa-auth-server/issues/881 here and making the button "Activate now" instead of just "Activate"?

templates/verify.txt
ryanfeeley: I don't recall the exclamation marks being called into question but **Congratulations** was a definite trigger.

johngruen: @ryanfeeley mailchip warns against "Excessive use of exclamation points" http://mailchimp.com/resources/guides/how-to-avoid-spam-filters/html/

Since the exclamation point here doesn't really add value, it might be better to err on the side of caution and get rid of it.

ryanfeeley: The exclamation adds a personal touch, and one is not excessive. A quick scan of verification emails from accounts this year include one is almost all of them somewhere in the email (including a new Gmail address). Let's keep it in.

johngruen:
@rfk done and done.

templates/verify.txt
johngruen: @ryanfeeley i sent an email earlier about spam testing, but for now, based on tests from my local build, the ! is a push. SHIPIT

https://youtu.be/--Vaz9jW054

vladikoff:
via maildev:

image

Thanks so much @johngruen !

PR #53, opened 2015-06-21
refactor(mailer): Extract boilerplate sending code into send.

shane-tomlinson:
@rfk - I promised it immediately. Here ya go.

This is running on stomlinson.dev.lcip.org


rfk:
LGTM!

PR #55, opened 2015-06-30
Append "now" to verify/reset password button CTAs

philbooth:
Fixes https://github.com/mozilla/fxa-auth-server/issues/881.

That issue only mentions the buttons, so I've not made the change in the text templates. Is that right/wrong?

### DIFF
--- a/templates/reset.html
+++ b/templates/reset.html
@@ -66,7 +66,7 @@
                                           <w:anchorlock/>
                                           <center>
                                             <![endif]-->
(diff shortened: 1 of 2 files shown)


rfk:
I think it makes sense to have "now" on the text templates as well, might as well keep them as consistent as possible.

philbooth:
Done.

rfk:
f+ from me, assigning to @ryanfeeley for confirmation from UX perspective

rfk:
After discussion with @shane-tomlinson let's hold off merging this until after train-41 is cut, so we don't introduce new strings.

rfk:
@philbooth the "verify" half of this was taken care of in john's refactor in #48, can you please rebase this to just the "reset" part?

philbooth:
@rfk rebased

PR #71, opened 2015-08-26
stubbed post_verify and reminder email text

### LINKED ISSUE #69: Prepare train-45 email string changes for translation
We have a suite of email refactorings lined up for train-45:

* #51 
* #60 
* #62 
* #65
* #68 

And maybe #66, although it has some open questions associated with it.

Many of these will require new of modified strings, so in order to give translators as much time as possible, let's try to land the strings first.  Following the pattern used by content-server, we can:

* make any trivial string changes directly in the existing templates
* put new strings that don't have a home yet in a `templates/strings.txt` where they can be found by the string-extraction script

We can then worry about the HTML, refactoring etc later on in the cycle.

@dannycoates can you please make this the first action on the email-quality bugs, to go through and find all the new strings and ensure they're captured appropriately?

dannycoates:
@ryanfeeley please activate... *er* verify these strings are correct as written.

Closes #69

### DIFF
--- a/templates/_pending.txt
+++ b/templates/_pending.txt
@@ -1,0 +1,11 @@
+# new string staging area
+
+# second reminder
+{{t "Still there?" }}
+{{t "A while ago you created a Firefox Account, but never verified it. Click the link below to activate your account." }}
+
+# post verify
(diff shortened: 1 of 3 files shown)


rfk:
@dannycoates I see what you did there...

Issue #68, opened 2015-07-27
Remove styles and link from verify email translation?


pdehaan:
Per [/templates/verify.html:55](https://github.com/mozilla/fxa-auth-mailer/blob/6aac195ff6d1692b239c7906fd1becd95ee85ebf/templates/verify.html#L55), we have the following string:

```
{{{t "For more information, please visit <a href='https://support.mozilla.org' style='color: #0095dd; text-decoration: none;'>Mozilla Support</a>"}}}
```

Not sure if we want to change the `href` URL to `%(supportLink)s` and the inline styles to `%(supportStyles)s` to make it harder for things to get altered or typoed.

shane-tomlinson:
> Not sure if we want to change the href URL to %(supportLink) and the inline styles to %(supportStyles) to make it harder for things to get altered or typoed.

I think this is a fantastic idea, if we can do it easily.

pdehaan:
@shane-tomlinson We do it already for some templates, for example, [templates/new_sync_device.html:57](https://github.com/mozilla/fxa-auth-mailer/blob/6aac195ff6d1692b239c7906fd1becd95ee85ebf/templates/new_sync_device.html#L57):

```
... immediately at <a href='%(resetLink)s'>%(resetLink)s</a>."}}}
```

In fact, here's my poor attempt at greppin' all `href=` tags that aren't parameterized:
```
➜  fxa-auth-mailer git:(master) grep -irn ' href=' templates
```

| FILE | STRING |
|----|----|
templates/new_sync_device.html:58: | `{{{t "For more information, please visit <a href='https://support.mozilla.org'>Mozilla Support</a>"}}}`
templates/password_changed.html:58: | `{{{t "For more information, please visit <a href='https://support.mozilla.org'>Mozilla Support</a>"}}}`
templates/password_reset.html:58: | `{{{t "For more information, please visit <a href='https://support.mozilla.org'>Mozilla Support</a>"}}}`
templates/reset.html:93: | `{{{t "For more information, please visit <a href='https://support.mozilla.org'>Mozilla Support</a>"}}}`
templates/unlock.html:91: | `{{{t "For more information, please visit <a href='https://support.mozilla.org'>Mozilla Support</a>"}}}`
templates/verify.html:55: | `{{{t "For more information, please visit <a href='https://support.mozilla.org' style='color: #0095dd; text-decoration: none;'>Mozilla Support</a>"}}}</p>`
```

rfk:
[...]

Issue #66, opened 2015-07-21
Send a post-verification welcome email


ryanfeeley:
To increase multiple device sync engagement, we should send users a welcome email that contains a sign-in link that, like our manage link, will sign them into their account (pre-fill the username field, or better, the cached page). We should measure the effectiveness of this email in terms of it increasing multi-device engagement.
![post-verification](https://cloud.githubusercontent.com/assets/68213/8805270/51858dd2-2f9e-11e5-84c8-9e012c8142d2.png)

rfk:
> We should measure the effectiveness of this email in terms of it increasing
> multi-device engagement.

We'll soon have explicit metrics on this front, so let's consider more precisely how we'll measure effectiveness of this.  Is it enough to say "after we started doing this, the rate of new multi-device connections per day increased from X to Y" or do you want to run a more formal able-driven experiment?

rfk:
> send users a welcome email that contains a sign-in link that [...] will sign them into their account

Hmmm...I assume we want this link to result in them being signed in to sync on their device.   Do we have the necessary magic in place to make that work on Android or iOS?  I know that https://www.mozilla.org/firefox/sync/ is capable of doing this on desktop by launching into about:accounts, but it doesn't seem to have the same capability on Android.

We may have to block on some or all of https://github.com/mozilla/fxa/issues/48 before we can implement this as described.

Or, we could ship an initial version that just links to https://www.mozilla.org/firefox/sync/ and then iterate on making things more automatic from there.

rfk:
@ryanfeeley we can try to ship an initial version of this on the coming train, but we'll need your input on precisely how to scope it.  Open questions:

* We don't have device names yet; can you re-word the "you successfully connected" part to avoid them for now?
* Where should all of the links go?
* What if the account was created when signing up for e.g. Pocket rather than sync?  Should we send a different email, the same, or no email at all?
* Does it make sense for us to send both this email, *and* the proposed JourneyBuilder emails from the engagement team?

ryanfeeley:
* RE: device names: You have successfully connected your first Firefox to your account.

* RE: Where should links go:
LINK (UTM parameters pendings)
* Android:
https://www.mozilla.org/firefox/android/?utm_source=firefox-browser&utm_medium=referral&utm_campaign=fxa-post-verification-email

* iOS: (the page will be live by Sept 1st, 2015 and it have a coming soon sign
up)
[...]

Issue #65, opened 2015-07-21
Send verification reminder emails


ryanfeeley:
When a user leaves an account unverified, everyone loses. We should send out two different verification reminders, and optimize when the send them out.

First Reminder – The earliest I would recommend we sent this would be 24 hours (and change "A few days ago" to "Recently"). Compare that against 48 and 72 hours?
![reminder1](https://cloud.githubusercontent.com/assets/68213/8803458/08139346-2f95-11e5-9d49-b6c48ecf1981.png)

Second Reminder – The earliest I would recommend we sent this would be 5 days (and change "A few weeks ago" to "A while ago"). Compare that against 8 and 13 days?
![reminder2](https://cloud.githubusercontent.com/assets/68213/8803459/08151f86-2f95-11e5-983a-42e03956c052.png)

We would want to find the most responded to day for either reminder and then use that.

Notice anything?
![915px-fibonacci_spiral_34 svg](https://cloud.githubusercontent.com/assets/68213/8803618/cae67398-2f95-11e5-8b7e-24f2fe01ac4b.png)

ryanfeeley:
@rfk Seems like low-effort high-impact based on the number of unverified accounts we have, no?

rfk:
Yeah, I've been treading water on this one a bit because the journeybuilder folks also wanted to take a crack at it, but we should just bite the bullet and manage this ourselves.

rfk:
@rfk to break out some sub-issues here

ckarlof:
We might need to double check with legal about this one. Transactional emails are "ok" from a legal POV, and it's not clear whether an email reminder for a previously transactional email is still considered transactional.

rfk:
I have vague recollections of getting a legal OK for this when the JourneyBuilder team was talking about doing it.  So I expect it will be OK, but we should definitely get confirmation.

dannycoates:
Is it worth the extra translation overhead to have 2 distinct sets of copy for the first and second reminder?

ckarlof:
For the record, I'm not warm on this one.

It's probably easy to do, but it also has the potential to annoy users if we do it badly. The upside is that it potentially gets us more accounts, but unfortunately only from users that weren't engaged enough to finish the process the begin with. 

[...]

Issue #62, opened 2015-07-15
Make all the emails look like the email format with the Firefox logo above


ryanfeeley:
I'm getting emails that don't look wonderful as the new email format. Please make them consistent. Unify?

rfk:
see also #49

ckarlof:
I'm closing #49, which is essentially a requirement how this bug should be implemented (i.e.. DRY).

rfk:
Partially addressed in https://github.com/mozilla/fxa-auth-mailer/pull/72, although we haven't yet split out the HTML into a re-useable template so I'm leaving this broader issue open for now.

PR #72, opened 2015-08-27
added post verify and reminder emails +housekeeping

dannycoates:
Sorry for jamming these all into one, but it made extracting strings easier.

#51
#60 
#62 
#68
#66 
#65

### DIFF
--- a/config.js
+++ b/config.js
@@ -10,10 +10,15 @@ module.exports = require('rc')(
       port: 9999,
       secure: false,
       sender: 'accounts@firefox.com',
-      verificationUrl: 'https://accounts.firefox.com/v1/verify_email',
-      passwordResetUrl: 'https://accounts.firefox.com/v1/complete_reset_password',
-      accountUnlockUrl: 'https://accounts.firefox.com/v1/complete_unlock_account',
-      initiatePasswordResetUrl: 'https://accounts.firefox.com/v1/reset_password'
+      verificationUrl: 'https://accounts.firefox.com/verify_email',
+      passwordResetUrl: 'https://accounts.firefox.com/complete_reset_password',
+      accountUnlockUrl: 'https://accounts.firefox.com/complete_unlock_account',
+      initiatePasswordResetUrl: 'https://accounts.firefox.com/reset_password',
+      initiatePasswordChangeUrl: 'https://accounts.firefox.com/settings/change_password',
+      syncUrl: 'https://www.mozilla.org/en-US/firefox/sync/',
+      androidUrl: 'https://www.mozilla.org/en-US/firefox/android/',
+      iosUrl: 'https://www.mozilla.org/en-US/firefox/ios/',
+      supportUrl: 'https://support.mozilla.org'
     }
   }
 )
--- a/mailer.js
+++ b/mailer.js
@@ -39,8 +39,13 @@ module.exports = function (log) {
     this.sender = config.sender
     this.verificationUrl = config.verificationUrl
     this.initiatePasswordResetUrl = config.initiatePasswordResetUrl
+    this.initiatePasswordChangeUrl = config.initiatePasswordChangeUrl
     this.passwordResetUrl = config.passwordResetUrl
     this.accountUnlockUrl = config.accountUnlockUrl
+    this.syncUrl = config.syncUrl
+    this.androidUrl = config.androidUrl
+    this.iosUrl = config.iosUrl
+    this.supportUrl = config.supportUrl
     this.translator = translator
     this.templates = templates
   }
@@ -117,7 +122,8 @@ module.exports = function (log) {
       templateValues: {
         email: message.email,
         link: link,
-        oneClickLink: oneClickLink
+        oneClickLink: oneClickLink,
+        supportUrl: this.supportUrl
       },
       uid: message.uid
     })
@@ -143,11 +149,12 @@ module.exports = function (log) {
         'X-Recovery-Code': message.code
       },
       subject: gettext('Reset your Firefox Account password'),
-      template: 'recoveryEmail',
+      template: 'resetEmail',
       templateValues: {
         code: message.code,
         email: message.email,
-        link: link
+        link: link,
+        supportUrl: this.supportUrl
       },
       uid: message.uid
     })
@@ -177,7 +184,8 @@ module.exports = function (log) {
       template: 'unlockEmail',
       templateValues: {
         email: message.email,
-        link: link
+        link: link,
+        supportUrl: this.supportUrl
       },
       uid: message.uid
     })
@@ -199,7 +207,8 @@ module.exports = function (log) {
       subject: gettext('Your Firefox Account password has been changed'),
       template: 'passwordChangedEmail',
       templateValues: {
-        resetLink: link
+        resetLink: link,
+        supportUrl: this.supportUrl
       },
       uid: message.uid
     })
@@ -218,7 +227,8 @@ module.exports = function (log) {
       subject: gettext('Your Firefox Account password has been reset'),
       template: 'passwordResetEmail',
       templateValues: {
-        resetLink: link
+        resetLink: link,
+        supportUrl: this.supportUrl
       },
       uid: message.uid
     })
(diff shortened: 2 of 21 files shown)


dannycoates:
Together with https://github.com/mozilla/fxa-auth-server/pull/1034 this is ready for review

config.js
```diff
@@ -13,7 +13,12 @@ module.exports = require('rc')(
       verificationUrl: 'https://accounts.firefox.com/v1/verify_email',
       passwordResetUrl: 'https://accounts.firefox.com/v1/complete_reset_password',
       accountUnlockUrl: 'https://accounts.firefox.com/v1/complete_unlock_account',
-      initiatePasswordResetUrl: 'https://accounts.firefox.com/v1/reset_password'
+      initiatePasswordResetUrl: 'https://accounts.firefox.com/reset_password',
```
rfk: I don't recall why these all have `v1` in there, but if we're dropping it from one, should we drop it from all?

config.js
```diff
@@ -13,7 +13,12 @@ module.exports = require('rc')(
...
+      initiatePasswordChangeUrl: 'https://accounts.firefox.com/settings/change_password',
+      syncUrl: 'https://www.mozilla.org/en-US/firefox/sync/',
+      androidUrl: 'https://www.mozilla.org/en-US/firefox/android/',
+      iosUrl: 'https://www.mozilla.org/en-US/firefox/ios/',
```
rfk: I wonder if we should name these with a little more detail about what they're for, e.g. "syncProductPageUrl".  Maybe I just have a thing about names..

mailer.js
```diff
@@ -226,10 +236,7 @@ module.exports = function (log) {
 
   Mailer.prototype.newSyncDeviceEmail = function (message) {
     log.trace({ op: 'mailer.newSyncDeviceEmail', email: message.email, uid: message.uid })
-    var query = {
-      email: message.email
-    }
-    var link = this.initiatePasswordResetUrl + '?' + qs.stringify(query)
+    var link = this.initiatePasswordChangeUrl
```
rfk: Does this new URL not take an `email` query param?  We should file a follow-up bug to add one if so, ISTM this is useful behaviour to coerce the site into the required state.

mailer.js
```diff
@@ -237,10 +244,70 @@ module.exports = function (log) {
       headers: {
         'X-Link': link
       },
-      subject: gettext('A new device is now syncing to your Firefox Account'),
+      subject: gettext('New sign-in to Sync'),
       template: 'newSyncDeviceEmail',
       templateValues: {
-        resetLink: link
+        resetLink: link,
```
rfk: This is no longer a "reset" link, should we change the name in the template to match?

mailer.js
```diff
@@ -237,10 +244,70 @@ module.exports = function (log) {
...
+        androidUrl: this.androidUrl,
+        iosUrl: this.iosUrl,
+        supportUrl: this.supportUrl
+      },
+      uid: message.uid
+    })
+  }
+
+  Mailer.prototype.reminderEmail = function (message) {
+    log.trace({ op: 'mailer.reminderEmail', email: message.email, uid: message.uid })
```
rfk: naming nit: reminder of what, or to do what?

templates/index.js
```diff
@@ -13,82 +13,64 @@ handlebars.registerHelper(
...
+module.exports = function () {
+  return P.all(
+    [
+      loadTemplates('verify'),
+      loadTemplates('reset'),
+      loadTemplates('unlock'),
+      loadTemplates('password_changed'),
+      loadTemplates('password_reset'),
+      loadTemplates('new_sync_device'),
+      loadTemplates('post_verify')
```
rfk: need to add "reminder" here?

rfk: This is a sign that we should add some tests, to at least assert we can correctly load and render each type of email; not necessarily in this PR though

rfk:
I don't feel well equipped to review all this HTML churn, will try loading them up in the browser and checking for consistency.

@dannycoates I think it got closed out in triage, but this really highlights the need for #49 to templatize this shared HTML.  Do you have a sense of how big a refactor that would be?  (It can safely be a follow-up bug).

templates/password_reset.html
```diff
@@ -55,7 +55,7 @@ <h1 style="text-align:center;margin-left:0;margin-bottom:10px;margin-right:0;mar
                           <tr>
                             <td valign="top" class="bodyContent" mc:edit="body_content" style="text-align:center;padding-left:20px;padding-bottom:30px;padding-right:20px;padding-top:30px;font-family:Helvetica;mso-table-rspace:0pt;mso-table-lspace:0pt;-ms-text-size-adjust:100%;-webkit-text-size-adjust:100%;color:#8A9BA8;line-height:13px;font-size:11px;">
                               {{{t "This is an automated email; if you didn't reset the password to your Firefox Account, you should reset it immediately at <a href='%(resetLink)s'>%(resetLink)s</a>."}}}
-                              {{{t "For more information, please visit <a href='https://support.mozilla.org'>Mozilla Support</a>"}}}
+                              {{{t "For more information, please visit <a href='%(supportUrl)s' style='color: #0095dd; text-decoration: none;'>Mozilla Support</a>"}}}
 
                             </td>
                           </tr>
```
rfk: Looks like `password_reset.html` hasn't been updated to the new format.

rfk:
> will try loading them up in the browser and checking for consistency

:+1: except for the password_reset template, which doesn't look like it's been updated to the new layout.

mailer.js
dannycoates: Good idea. I don't think content-server does anything with it on this page yet, but there's no harm in adding it here so we could autofill the login form if they aren't already logged in.

mailer.js
dannycoates: I didn't want to stomp on more translation strings than we already have for an "invisible" change.

templates/index.js
dannycoates: I left it out intentionally, but since you mention it, not for a good reason. This repo needs more tests for sure though.

mailer.js
dannycoates: Part of me says, when we send more than one kind I'll care, the other part capitulates to your infernally sound logic.

config.js
dannycoates: I'll give you `reminder` but you've gone too [Objective-C](https://www.quora.com/What-is-the-longest-Objective-C-method-name-in-Cocoa) on these

dannycoates:
> Do you have a sense of how big a refactor that would be?

Not too big, but there were enough little differences that it was still way easier just to copy. I feel like the next UI change should be the one that breaks it up into a template with partials.

config.js
rfk: See, that's why we make such a good team ;-)

rfk:
This LGTM; @dannycoates is it ready for final review and merge?

rfk:
> I feel like the next UI change should be the one that breaks it up into a template with partials

Let's leave #62 open until we get around to this then.

dannycoates:
> is it ready for final review and merge?

yes

templates/_pending.txt
```diff
@@ -5,7 +5,7 @@
 {{t "A while ago you created a Firefox Account, but never verified it. Click the link below to activate your account." }}
 
 # post verify
-{{t "You successfully connected your first device to your Firefox Account. Now you can sign in to Sync from your other devices, including Firefox for <a href='%(androidLink)s'>Android</a> and <a href='%(iosLink)s'>iOS</a>." }}
-{{t "You successfully connected your first device to your Firefox Account. To make the most of Sync, connect any <a href='%(androidLink)s'>Firefox for Android</a> or <a href='%(iosLink)s'>Firefox for iOS</a>."}}
+{{t "You successfully connected your first device to your Firefox Account. Now you can sign in to Sync from your other devices, including Firefox for <a href='%(androidUrl)s'>Android</a> and <a href='%(iosUrl)s'>iOS</a>." }}
+{{t "You successfully connected your first device to your Firefox Account. To make the most of Sync, connect any <a href='%(androidUrl)s'>Firefox for Android</a> or <a href='%(iosUrl)s'>Firefox for iOS</a>."}}
```
rfk: Do these string changes require an explicit update to https://github.com/mozilla/fxa-content-server-l10n/pull/68/ before it gets merged?

rfk:
LGTM, thanks @dannycoates!

templates/_pending.txt
dannycoates: already done

ckarlof:
@rfk @ryanfeeley I feel dumb, but it's been a while since I've signed up for an account and saw these. Awesome! Do we have any data on how effective they are in creating multi-device usage?

rfk:
>  Do we have any data on how effective they are in creating multi-device usage?

I don't know of any specifically.  I feel like we had an open issue to add utm_* params to these links but I can't see it now, I'll open a fresh one.

rfk:
Oh, wait, it's here, fairly recently: https://github.com/mozilla/fxa-auth-mailer/pull/118/files

PR #87, opened 2015-09-28
fix(templates): remove extra alt text from the Firefox logo

### LINKED ISSUE #86: Alt tag for Firefox logo in the email should be "Firefox logo"
Low priority, but easy to do, the alt tag for Firefox logo in the email should be "Firefox logo" and not just "Firefox". Seeing a lot of "Firefox" in user testing when people check their webmail and images aren't loading.

vladikoff:
Fixes #86


vladikoff:
@shane-tomlinson @zaach r?

PR #89, opened 2015-09-29
fix(templates): share the same logo template with all templates, clea…

### LINKED ISSUE #62: Make all the emails look like the email format with the Firefox logo above

vladikoff:
…n up HTML, fix font bugs

Fixes #62


vladikoff:
Testing with email on acid...

vladikoff:
@rfk r? 

Email on acid results: https://www.emailonacid.com/app/acidtest/display/summary/N9ui2lkD4g9cvmFz1L1eYKFEXxuVlKZTCjHB49qGCQWi3

This fixes several Outlook issues but breaks old Lotus < 8 notes. Let me know if we should care about those versions.

rfk:
> but breaks old Lotus < 8 notes. Let me know if we should care about those versions. 

Nope.

It would be awesome to fix this everywhere, but I'll happily trade off looking worse in Lotus Notes for looking better in Outlook.

package.json
```diff
@@ -18,6 +18,7 @@
     "bluebird": "2.9.34",
     "bunyan": "1.0.0",
     "fxa-content-server-l10n": "git://github.com/mozilla/fxa-content-server-l10n.git",
+    "grunt-nunjucks-2-html": "vitkarpov/grunt-nunjucks-2-html#1900f91a756b2eaf900b20",
```
rfk: Out of curiosity, is nunjucks better suited for this than handlebars, which we already have as a dependency?

README.md
```diff
@@ -15,6 +15,12 @@ The auth-mailer also includes a restify API to send emails, but the auth server
 * npm
 * postfix
 
+## Changing Templates
+
+If you are changing or adding templates then you need to update `.html` and `.txt` templates.
+Use the `/partials` directory to make changes to the HTML templates, then run `grunt templates` to regenerate the template.
+This saves the HTML template into `/templates`. Then make changes to the `.txt` template in the `/templates` directory.
```
rfk: As a follow-up bug, perhaps we can automatically validate this with a pre-commit hook that checks modified times on the relevant file pairs.

partials/new_sync_device.html
```diff
@@ -0,0 +1,18 @@
+<% extends "partials/parent.html" %>
+
+<% block content %>
+<!--Header Area-->
+<tr style="page-break-before: always">
```
rfk: Is `page-break-before: always` the magic sauce that prevents outlook breakage?

templates/README.md
```diff
@@ -0,0 +1 @@
+HTML templates in this directory are dynamically generated using the 'grunt nunjucks' task....
```
rfk: nit: IIUC `grunt templates` is the preferred way of calling it

templates/parent.html
```diff
@@ -0,0 +1,25 @@
+<!DOCTYPE html PUBLIC "-//W3C//DTD XHTML 1.0 Transitional//EN" "http://www.w3.org/TR/xhtml1/DTD/xhtml1-transitional.dtd">
```
rfk: It's harmless, but is there a simple way to prevent the grunt task for generating this stub "parent" template?

rfk:
Thanks @vladikoff, this is great cleanup.  A couple of nits and q's but r+ with or without them addressed.

I feel like we can iterate towards another level of abstraction with the templating.  For example AFAICT the only bits that differ between `new_sync_device.html` and `password_changed.html` are the text strings, so maybe we can generate both `.html` and `.txt` versions from a single source that just specifies those strings.  But that can be for a separate bug.

package.json
vladikoff: Handlebars does not support layout blocks, while Nunjucks, Jade, Jinja templating do

vladikoff:
>> but breaks old Lotus < 8 notes. Let me know if we should care about those versions.
Nope.

> It would be awesome to fix this everywhere, but I'll happily trade off looking worse in Lotus Notes for looking better in Outlook.

So I checked it again and basically the long "Alternatively" link is what breaks lotus notes, I guess it was always broken but we didn't test with the long link before, to the changes in this PR did not affect that behaviour

partials/new_sync_device.html
vladikoff: Yeah and it also helps to split the table into 2

package.json
rfk: SGTM

rfk:
assigning this back to @vladikoff for nits and merge

package.json
shane-tomlinson: YES! I would love to move the content server to Nunjucks, but we are pretty far down the Handlebars/Mustache hole now.

vladikoff:
Test results: 

https://www.emailonacid.com/app/acidtest/display/summary/UOKEA2Hr0fSVC95ymfXikGVtccwaumNeO1gVJNwOAyKnZ

https://www.emailonacid.com/app/acidtest/display/summary/ZR5ZNrk0YXIiVb6QP11DarxJMrkIxf3Na87LmssYII3bM

https://www.emailonacid.com/app/acidtest/display/summary/KUoXiakauKoBXjkvb9djMMlFM1swcFuCMX37KxYOFP0s3

https://www.emailonacid.com/app/acidtest/display/summary/pc5aPfOZEY5AwhB4gydsDXMQv04wZtJktWeQDHgz7DYZD

PR #92, opened 2015-10-05
fix(l10n): remove html attributes from string

### LINKED ISSUE #74: Issues with train 45 strings
From @mathjazz in https://github.com/mozilla/fxa-content-server-l10n/pull/68:
```
I'd suggest against putting this into localizable strings, because it's confusing localizers, some of them will translate these words:
<a href='%(supportUrl)s' style='color: #0095dd; text-decoration: none;'>

Also, do we really need all these string:

Activate Now
Activate now
Activate now:
```

zaach:
Fixes #74 
@dannycoates r?

When merging the next batch of extracted strings, we'll need to convert/regex-foo the old stings into this new format so we don't lose any translations.

### DIFF
--- a/mailer.js
+++ b/mailer.js
@@ -54,6 +54,12 @@ module.exports = function (log) {
     this.mailer.close()
   }
 
+  Mailer.prototype._supportLinkAttributes = function () {
+    // Not very nice to have presentation code in here, but this is to help l10n
+    // contributors not deal with extraneous noise in strings.
+    return 'href="' + this.supportUrl + '" style="color: #0095dd; text-decoration: none;"'
+  }
+
   Mailer.prototype.send = function (message) {
     log.trace({ op: 'mailer.' + message.template, email: message.email, uid: message.uid })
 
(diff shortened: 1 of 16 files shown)


vladikoff:
@zaach see https://github.com/mozilla/fxa-auth-mailer#changing-templates

zaach:
@shane-tomlinson should we take care of this for the upcoming extraction or wait for the next one?

shane-tomlinson:
@zaach - I was at the point of merging this until I realized it removes the href attribute from the links. That change will cause all non-updated templates to fail the lint task, because we now check all element attributes against a list of known good attributes. Since we are already late cutting the strings, let's wait for the next train to give localizers 2 full weekends.

shane-tomlinson:
Assigning to me to merge as soon as train 47 is cut.

zaach:
@shane-tomlinson we can run a search/replace on all of the PO files after we merge this, that way localizers won't have to retranslate it as the only thing that changed was inside the `a` tag.

PR #93, opened 2015-10-06
fix(email): add better password change link for certain emails

### LINKED ISSUE #78: "New sign-in to Sync" mail has bad advice re ignoring the email if received in error.
The fine-print in the mail says "This is an automated email; if you received it in error, no action is required." - I'm not sure that is good advice as it implies someone has their password.

vladikoff:
Fix #78

### DIFF
--- a/mailer.js
+++ b/mailer.js
@@ -57,7 +57,13 @@ module.exports = function (log) {
   Mailer.prototype._supportLinkAttributes = function () {
     // Not very nice to have presentation code in here, but this is to help l10n
     // contributors not deal with extraneous noise in strings.
-    return 'href="' + this.supportUrl + '" style="color: #0095dd; text-decoration: none;"'
+    return 'href="' + this.supportUrl + '" style="color: #0095dd; text-decoration: none; font-family: sans-serif;"'
+  }
+
+  Mailer.prototype._initiatePasswordChange = function () {
+    // Not very nice to have presentation code in here, but this is to help l10n
+    // contributors not deal with extraneous noise in strings.
+    return 'href="' + this.initiatePasswordChangeUrl + '" style="color: #0095dd; text-decoration: none; font-family: sans-serif;"'
   }
 
   Mailer.prototype.send = function (message) {
@@ -130,7 +136,8 @@ module.exports = function (log) {
         link: link,
         oneClickLink: oneClickLink,
         supportUrl: this.supportUrl,
-        supportLinkAttributes: this._supportLinkAttributes()
+        supportLinkAttributes: this._supportLinkAttributes(),
+        passwordChangeLinkAttributes: this._initiatePasswordChange()
       },
       uid: message.uid
     })
@@ -162,7 +169,8 @@ module.exports = function (log) {
         email: message.email,
         link: link,
         supportUrl: this.supportUrl,
-        supportLinkAttributes: this._supportLinkAttributes()
+        supportLinkAttributes: this._supportLinkAttributes(),
+        passwordChangeLinkAttributes: this._initiatePasswordChange()
       },
       uid: message.uid
     })
@@ -194,7 +202,8 @@ module.exports = function (log) {
         email: message.email,
         link: link,
         supportUrl: this.supportUrl,
-        supportLinkAttributes: this._supportLinkAttributes()
+        supportLinkAttributes: this._supportLinkAttributes(),
+        passwordChangeLinkAttributes: this._initiatePasswordChange()
       },
       uid: message.uid
     })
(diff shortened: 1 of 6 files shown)


vladikoff:
still wip, 

- need to check if we want this new link to /settings (as suggested in https://github.com/mozilla/fxa-auth-mailer/issues/78)
- need to check if this should be added in other emails besides just the new sync device

vladikoff:
@zaach @rfk r?

mailer.js
```diff
@@ -162,7 +169,8 @@ module.exports = function (log) {
         email: message.email,
         link: link,
         supportUrl: this.supportUrl,
-        supportLinkAttributes: this._supportLinkAttributes()
+        supportLinkAttributes: this._supportLinkAttributes(),
+        passwordChangeUrl: this._initiatePasswordChange()
```
rfk: naming nit: `passwordChangeLinkAttributes` or similar for consistency, since it's not just the URL?

mailer.js
```diff
@@ -321,7 +334,8 @@ module.exports = function (log) {
         link: link,
         oneClickLink: oneClickLink,
         supportUrl: this.supportUrl,
-        supportLinkAttributes: this._supportLinkAttributes()
+        supportLinkAttributes: this._supportLinkAttributes(),
+        passwordChangeUrl: this._initiatePasswordChange()
```
rfk: Wow that's a lot of identical changes...I wonder if we should factor these links out into some common set of attributes that are included in every render, so that we can manage them in a single place.

rfk:
@vladikoff travis bustage?

rfk:
Also if I understand the rules correctly (thanks to a recent refresher from @shane-tomlinson) then we shouldn't merge this l10n-affecting string change until after we cut the next train...?

vladikoff:
will fix the tests and ping @shane-tomlinson before cutting strings....

shane-tomlinson:
> Also if I understand the rules correctly (thanks to a recent refresher from @shane-tomlinson) then we shouldn't merge this l10n-affecting string change until after we cut the next train...?

We are in the "merge new strings!" window, until I go to sleep tonight.

mailer.js
vladikoff: opened https://github.com/mozilla/fxa-auth-mailer/issues/94 so we can merge this and cut strings

test/local/mailer_tests.js
```diff
@@ -43,7 +43,7 @@ P.all(
           service: 'service',
         }
 
-        var supportHtmlLink = new RegExp('<a href="' + config.mail.supportUrl + '" style="color: #0095dd; text-decoration: none;">Mozilla Support</a>')
+        var supportHtmlLink = new RegExp('<a href="' + config.mail.supportUrl + '" style="color: #0095dd; text-decoration: none; font-family: sans-serif;">Mozilla Support</a>')
```
shane-tomlinson: Should we add a "Change Password" link test too?

templates/password_reset.html
```diff
@@ -27,7 +27,7 @@ <h1 style="font-family: sans-serif; font-weight: normal; margin: 0 0 24px 0; tex
   <td border="0" cellpadding="0" cellspacing="0" height="100%" width="100%">
     <br/>
     <p class="secondary" style="font-family: sans-serif; font-weight: normal; margin: 0; text-align: center; color: #8A9BA8; font-size: 11px; line-height: 13px; width: 310px !important; word-wrap:break-word">
-      {{t "This is an automated email; if you received it in error, no action is required."}} {{{t "For more information, please visit <a %(supportLinkAttributes)s>Mozilla Support</a>"}}}
+      {{t "This is an automated email; if you did not authorize this action, then <a %(passwordChangeLinkAttributes)s>please change your password.</a>" }} {{{t "For more information, please visit <a %(supportLinkAttributes)s>Mozilla Support</a>"}}}
```
shane-tomlinson: Ahh, since this now has HTML, you'll have to replace `{{` with `{{{`or else the HTML will be escaped.

test/local/mailer_tests.js
vladikoff: added to https://github.com/mozilla/fxa-auth-mailer/issues/94

templates/password_reset.html
shane-tomlinson: That's my theory anyways.

mailer.js
```diff
@@ -130,7 +136,8 @@ module.exports = function (log) {
         link: link,
         oneClickLink: oneClickLink,
         supportUrl: this.supportUrl,
-        supportLinkAttributes: this._supportLinkAttributes()
+        supportLinkAttributes: this._supportLinkAttributes(),
+        passwordChangeLinkAttributes: this._initiatePasswordChange()
```
shane-tomlinson: Should `_initiatePasswordChange` be renamed to `_passwordChangeLinkAttributes` to stay consistent?

templates/password_reset.html
vladikoff: yea good note fixed via https://github.com/mozilla/fxa-auth-mailer/commit/83984f6789b8e998a1fea2a3907fa8ab40a9a2bd

mailer.js
vladikoff: added a note to https://github.com/mozilla/fxa-auth-mailer/issues/94#issuecomment-148462202

PR #97, opened 2015-10-19
fix(email): disable new sync device email

### LINKED ISSUE #96: Discontinue new Sync device email
Our activity email contains very little information and is likely scaring users.

https://support.mozilla.org/en-US/questions/1089471

Let's stop sending it for now. It doesn't really offer any value until we can share more information about the new device, etc.

vladikoff:
Fixes #96

### DIFF
--- a/mailer.js
+++ b/mailer.js
@@ -257,25 +257,7 @@ module.exports = function (log) {
   }
 
   Mailer.prototype.newSyncDeviceEmail = function (message) {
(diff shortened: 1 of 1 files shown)


vladikoff:
> ]  <~rfkelly>	it may be simpler to just delete the call to newSyncDeviceEmail in fxa-auth-server

Doing that instead

PR #102, opened 2015-11-11
fix(config): remove rc

### LINKED ISSUE #18: Switch from "rc" to "convict" for config management
I seem to recall us doing an rc => convict switch for other fxa-* projects, but this one didn't receive the treatment.

vladikoff:
Fixes #18


vladikoff:
@rfk r?

config.js
```diff
@@ -1,24 +1,140 @@
...
+    env: 'NODE_ENV'
+  },
+  port: {
+    env: 'MAILER_PORT',
+    format: 'port',
+    default: 1810
+  },
+  logging: {
+    app: {
+      default: 'fxa-profile-server'
```
rfk: copypasta "fxa-profile-server" should be "fxa-auth-mailer"

config.js
```diff
@@ -1,24 +1,140 @@
...
+    },
+    androidUrl: {
+      doc: 'url to Android product page',
+      format: String,
+      default: 'https://www.mozilla.org/en-US/firefox/android/'
+    },
+    iosUrl: {
+      doc: 'url to IOS product page',
+      format: String,
+      default: 'https://www.mozilla.org/en-US/firefox/ios/'
```
rfk: It's not related to your changes, but I noticed it while I was in here: we shouldn't embed locale information in these URLs, should use e.g. 'https://www.mozilla.org/firefox/ios' and let the server figure out the right locale at runtime.  May as well fix while we're in here.

test/mail_helper.js
```diff
@@ -1,101 +0,0 @@
-/* This Source Code Form is subject to the terms of the Mozilla Public
- * License, v. 2.0. If a copy of the MPL was not distributed with this
- * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
-
-var MailParser = require('mailparser').MailParser
-var restify = require('restify')
-var config = require('rc')(
-  'fxa_mail_helper',
```
rfk: Huh, was this leftover code coped from fxa-auth-server that isn't used for the tests in this repository?

rfk:
r+ module nits, thanks @vladikoff.

Oh, one more it, I suggest renaming the commit to something like "replace rc with convict" rather than just "remove rc" :-)

test/mail_helper.js
vladikoff: exactly, older version of it too, it seems

config.js
vladikoff: going to fix that

vladikoff:
Updated and addressed feedback, thanks Ryan!

PR #106, opened 2015-12-15
fix(deps): update dependencies

vladikoff:
@rfk r?

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,7 +4,7 @@
   "description": "Library to send out verification emails in the fxa-auth-server which renders emails from a template (and handles localization).",
   "main": "index.js",
   "scripts": {
-    "test": "tap test/local",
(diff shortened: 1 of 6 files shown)


package.json
```diff
@@ -16,9 +16,9 @@
   "bugs": "https://github.com/mozilla/fxa-auth-mailer/issues",
   "license": "MPL-2.0",
   "dependencies": {
-    "bluebird": "2.9.34",
+    "bluebird": "3.0.6",
```
rfk: Is this the bluebird version of death that caused a memory-leak in the auth-server?  It may be worth holding off on upgrading that across the board until we understand that's going on there (unless it's a security update of course).

rfk:
otherwise :+1:

vladikoff:
Thanks for the review @rfk !

PR #107, opened 2015-12-30
feat(log): switch to mozlog

vladikoff:


vladikoff:
@rfk r?

Extracting this from the reminder PR to simplify changes

bin/server.js
```diff
@@ -4,11 +4,15 @@
...
+
+var log = require('../log')('server')
+var mailerLog = require('../log')('mailer')
+var mailConfig = config.get('mail')
+
 var packageJson = require('../package.json')
 var P = require('bluebird')
-var Mailer = require('../mailer')(log)
+// NOTE: Mailer is also used by fxa-auth-server directly, the module has two different logging interfaces
+var Mailer = require('../mailer')(mailerLog)
```
rfk: For readability, it may make sense to move the declaration of `mailerLog` down so that it's underneath the comment as well.

config.js
```diff
@@ -124,7 +124,7 @@ const conf = convict({
   }
 })
 
-var envConfig = path.join(__dirname, '..', 'config', conf.get('env') + '.json')
+var envConfig = path.join(__dirname, 'config', conf.get('env') + '.json')
```
rfk: Was this a straight-up bug in the previous version, looking for the files in the wrong place?

rfk:
There are several logging calls in mailer.js like this:

```
    log.info({ op: 'mailer.send', email: message.email })
```

Should they also be refactored to use the mozlog form? Like:

```
    log.info('send', { email: message.email })
```

We'll also need to check with @jrgm whether this changes the name or format of any logs currently being processed for monitoring/alerting purposes.

rfk:
Also, \o/ consistency! :-)

config.js
vladikoff: Yeah this was broken

vladikoff:
> There are several logging calls in mailer.js like this:

>    log.info({ op: 'mailer.send', email: message.email })
Should they also be refactored to use the mozlog form? Like:

>    log.info('send', { email: message.email })
We'll also need to check with @jrgm whether this changes the name or format of any logs currently being processed for monitoring/alerting purposes.

I keep coming back to this, I think I will have to update this PR to abstract the old logging in the same way : https://github.com/mozilla/fxa-auth-server/blob/master/lib/log.js#L40

PR #113, opened 2016-02-03
feat(db): add ability to check account verification status

### LINKED ISSUE #111: add ability to talk to fxa-db
We need ability to talk to the fxa db to check user's verification status.

vladikoff:
Fixes #111

### DIFF
--- a/config.js
+++ b/config.js
@@ -19,6 +19,19 @@ var conf = convict({
     format: 'port',
     default: 10136
   },
+  db: {
+    backend: {
+      default: 'httpdb',
+      env: 'DB_BACKEND'
+    }
+  },
+  httpdb: {
+    url: {
+      doc: 'database api url',
+      default: 'http://127.0.0.1:8000',
+      env: 'HTTPDB_URL'
+    }
+  },
   logging: {
     app: {
       default: 'fxa-auth-mailer'
--- a/lib/crypto/butil.js
+++ b/lib/crypto/butil.js
@@ -1,0 +1,36 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+var HEX = /^(?:[a-fA-F0-9]{2})+$/
+
+module.exports.unbuffer = function unbuffer(object, inplace) {
+  var keys = Object.keys(object)
+  var copy = inplace ? object : {}
+  for (var i = 0; i < keys.length; i++) {
+    var x = object[keys[i]]
+    copy[keys[i]] = Buffer.isBuffer(x) ? x.toString('hex') : x
+  }
+  return copy
+}
+
+module.exports.bufferize = function bufferize(object, options) {
+  var keys = Object.keys(object)
+  options = options || {}
+  var copy = options.inplace ? object : {}
+  var ignore = options.ignore || []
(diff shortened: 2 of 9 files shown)


lib/crypto/butil.js
```diff
@@ -0,0 +1,62 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
```
vladikoff: I wonder if we should make this a module, it is the same as in fxa-auth-server

cc @rfk

config.js
```diff
@@ -19,6 +19,19 @@ const conf = convict({
...
+  db: {
+    backend: {
+      default: 'httpdb',
+      env: 'DB_BACKEND'
+    }
+  },
+  httpdb: {
+    url: {
+      doc: 'database api url',
+      default: 'http://127.0.0.1:8000',
```
vladikoff: TODO: this might need to be changed

lib/crypto/butil.js
rfk: lol, I was just privately thinking "I'm not thrilled about all this code duplication from the auth-server" :-)

package.json
```diff
@@ -24,13 +24,17 @@
     "handlebars": "1.3.0",
     "i18n-abide": "0.0.23",
     "jed": "0.5.4",
+    "joi": "6.9.1",
```
vladikoff: TODO: remove

lib/crypto/butil.js
vladikoff: @jrm says @seanmonstar also has a buffer util module, need to check that

seanmonstar: https://www.npmjs.com/package/buf

Allows me to do `db.getToken(buf(token))` that works with either a hex string of a buffer, and then I can `reply({ token: hex(token) })` for the reverse: give me a hex string if its a Buffer, but don't error if it's already a string.

lib/crypto/butil.js
```diff
@@ -0,0 +1,62 @@
...
+var HEX = /^(?:[a-fA-F0-9]{2})+$/
+
+module.exports.buffersAreEqual = function buffersAreEqual(buffer1, buffer2) {
```
seanmonstar: PS: `Buffer.equals` exists starting with nodejs 0.12. `bufOne.equals(bufTwo)` https://nodejs.org/dist/latest-v4.x/docs/api/buffer.html#buffer_buf_equals_otherbuffer

test/local/db_tests.js
```diff
@@ -0,0 +1,107 @@
...
+        t.pass('Got the ping ok')
+      }, function (err) {
+        throw err
+      })
+  }
+)
+
+
+test(
+  'get email record',
```
vbudhram: The `pool.js` file exposes alot of functionality. Is it worthwhile to test other HTTP Methods for email?

lib/crypto/butil.js
vladikoff: Ok thanks, looks like the `bufferize` in fxa-auth-server works on Objects, so I'm gonna keep that one for now.

test/local/db_tests.js
vladikoff: :+1:

vladikoff:
@vbudhram updated, could you do another r? when you have time

vbudhram:
@vladikoff LGTM

PR #117, opened 2016-02-05
chore(travis): run grunt l10n-extract on travis

### LINKED ISSUE #116: Run `grunt l10n-extract` on Travis-CI
> Not sure if we want to tweak the [.travis.yml](https://github.com/mozilla/fxa-auth-mailer/blob/26ced35d881466fb4b38b093445b18c0c2db9281/.travis.yml#L33) file to run the Grunt `default` task (instead of just `grunt templates`), so it also tests `l10n-extract` on each commit.
> &mdash; via https://github.com/mozilla/fxa-auth-mailer/issues/114#issuecomment-179969634

Or not, but would help catch future potential issues a la #114

vladikoff:
Fixes #116

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -29,6 +29,7 @@ install:
   - if [ $TRAVIS_NODE_VERSION == "4" ]; then CXX=g++-4.8 npm install; else npm install; fi
 
 script:
-  - npm outdated --depth 0
   - grunt templates
   - npm test
+  - npm outdated --depth 0
+  - grunt l10n-extract


.travis.yml
```diff
@@ -29,6 +29,7 @@ install:
   - if [ $TRAVIS_NODE_VERSION == "4" ]; then CXX=g++-4.8 npm install; else npm install; fi
 
 script:
-  - npm outdated --depth 0
   - grunt templates
   - npm test
+  - npm outdated --depth 0
+  - grunt l10n-extract
```
pdehaan: You could possibly just call `grunt` and use the default task which calls both `grunt templates` and `grunt l10n-extract`.

vladikoff: I kinda feel like `l10n-extract` lives in its own world (string extraction, etc). It is not essential for the project build.

PR #126, opened 2016-03-11
fix(config): add content server config values

vladikoff:
Should match: https://github.com/mozilla/fxa-auth-server/blob/master/config/index.js#L371

@rfk @vbudhram r?

### DIFF
--- a/config.js
+++ b/config.js
@@ -32,6 +32,13 @@ var conf = convict({
       env: 'HTTPDB_URL'
     }
   },
+  contentServer: {
+    url: {
+      doc: 'The url of the corresponding fxa-content-server instance',
+      default: 'http://127.0.0.1:3030',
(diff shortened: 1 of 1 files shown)


config.js
```diff
@@ -32,6 +32,13 @@ var conf = convict({
       env: 'HTTPDB_URL'
     }
   },
+  contentServer: {
+    url: {
+      doc: 'The url of the corresponding fxa-content-server instance',
+      default: 'https://accounts.firefox.com',
```
rfk: nit: the auth-server uses ` default: 'http://127.0.0.1:3030'` for this, may be worth doing the same for full consistency.  I don't feel strongly either way.

rfk:
Looks good, r+ with one tiny nit

PR #132, opened 2016-04-01
fix(templates): make base partial responsive

ckarlof:
These are changes suggested by Jessilyn to the base template to make it responsive. It's been roughly tested in Email on Acid, but probably needs more.

### DIFF
--- a/partials/base/base.html
+++ b/partials/base/base.html
@@ -3,10 +3,17 @@
 <head>
   <meta http-equiv="Content-Type" content="text/html; charset=UTF-8" />
   <title>{{t "Firefox Accounts"}}</title>
+  <style type="text/css">
+   @media only screen and (max-width: 480px) {
+     div[class="mobiledevicesupport"] table.container {
+       width: 320px !important;
+     }
+   }
+	 </style>
 </head>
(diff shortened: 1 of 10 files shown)


partials/base/base.html
```diff
@@ -3,10 +3,17 @@
...
+  <style type="text/css">
+   @media only screen and (max-width: 480px) {
+     div[class="mobiledevicesupport"] table.container {
+       width: 320px !important;
+     }
+   }
+	 </style>
 </head>
-
+<div class="mobiledevicesupport">
```
vladikoff: @ckarlof a div outside the body does not seem right, should we put that into the body?

ckarlof:
There some centering problems with the changes.

partials/base/base.html
ckarlof: lol, indeed. These changes were suggested by jdavis from the email team late on Friday, and I was mainly just trying to capture and distribute them to the wider team.

vladikoff:
@ryanfeeley our emails already fit well in mobile: 



Let me know if setting a 480px media query is necessary...

PR #137, opened 2016-04-06
fix(server): Fix the port binding.

### LINKED ISSUE #136: Server does not bind to port defined in config
When I start the server and do an `lsof -i :10136`, no result is returned. 

If I change [this line](https://github.com/mozilla/fxa-auth-mailer/blob/d74a9e5c4ee33e1dd0b84c62871fda604c71cb5e/bin/server.js#L73) to read:

```
 log.info('listening', api.url)
```

When the server starts, it shows: `listening http://[::]:63596`.

The port is not defined.  [`config.port`](https://github.com/mozilla/fxa-auth-mailer/blob/d74a9e5c4ee33e1dd0b84c62871fda604c71cb5e/bin/server.js#L71) should be `config.get('port')`.

shane-tomlinson:
Fetch the port from config using `config.get('port')` instead of `config.port`.

fixes #136

### DIFF
--- a/bin/server.js
+++ b/bin/server.js
@@ -68,7 +68,7 @@ P.all(
     )
 
     api.listen(
-      config.port,
+      config.get('port'),
       function () {
         log.info('listening', { port: config.get('port') })
       }


vladikoff:
Yeap!

PR #138, opened 2016-04-08
feat(mailer): Add new security templates

shane-tomlinson:
New templates:
* suspicious location
* reset_password_required

Add the ability to pass in a "sender" to the Mailer, for testing.


shane-tomlinson:
The localized strings for this PR are at https://github.com/mozilla/fxa-content-server-l10n/pull/113.

PR #141, opened 2016-04-11
feat(security): signin confirmation email

shane-tomlinson:
This PR contains the re-verify email template. Based on the original verify email template.

* [x] Better verbage on email
* [x] String extraction

### DIFF
--- a/config.js
+++ b/config.js
@@ -95,6 +95,11 @@ var conf = convict({
       format: String,
       default: undefined
     },
+    verifyLoginUrl: {
+      doc: 'Deprecated. uses contentServer.url',
+      format: String,
+      default: undefined
+    },
     passwordResetUrl: {
       doc: 'Deprecated. uses contentServer.url',
       format: String,
@@ -156,11 +161,13 @@ var options = {
 
 conf.validate(options)
 
-conf.set('mail.signInUrl', conf.get('contentServer.url') + '/signin')
-conf.set('mail.verificationUrl', conf.get('contentServer.url') + '/v1/verify_email')
-conf.set('mail.passwordResetUrl', conf.get('contentServer.url') + '/v1/complete_reset_password')
-conf.set('mail.accountUnlockUrl', conf.get('contentServer.url') + '/v1/complete_unlock_account')
-conf.set('mail.initiatePasswordResetUrl', conf.get('contentServer.url') + '/reset_password')
-conf.set('mail.initiatePasswordChangeUrl', conf.get('contentServer.url') + '/settings/change_password')
+var contentServerUrl = conf.get('contentServer.url')
+conf.set('mail.signInUrl', contentServerUrl + '/signin')
+conf.set('mail.verificationUrl', contentServerUrl + '/v1/verify_email')
+conf.set('mail.passwordResetUrl', contentServerUrl + '/v1/complete_reset_password')
+conf.set('mail.accountUnlockUrl', contentServerUrl + '/v1/complete_unlock_account')
+conf.set('mail.initiatePasswordResetUrl', contentServerUrl + '/reset_password')
+conf.set('mail.initiatePasswordChangeUrl', contentServerUrl + '/settings/change_password')
+conf.set('mail.verifyLoginUrl', contentServerUrl + '/complete_signin')
 
 module.exports = conf
--- a/mailer.js
+++ b/mailer.js
@@ -44,6 +44,7 @@ module.exports = function (log) {
     this.mailer = sender || nodemailer.createTransport('SMTP', options)
     this.sender = config.sender
     this.verificationUrl = config.verificationUrl
+    this.verifyLoginUrl = config.verifyLoginUrl
     this.initiatePasswordResetUrl = config.initiatePasswordResetUrl
     this.initiatePasswordChangeUrl = config.initiatePasswordChangeUrl
     this.passwordResetUrl = config.passwordResetUrl
@@ -143,6 +144,8 @@ module.exports = function (log) {
   }
 
   Mailer.prototype.verifyEmail = function (message) {
+    log.trace({ op: 'mailer.verifyEmail', email: message.email, uid: message.uid })
+
     var query = {
       uid: message.uid,
       code: message.code
@@ -174,6 +177,45 @@ module.exports = function (log) {
         supportUrl: this.supportUrl,
         supportLinkAttributes: this._supportLinkAttributes(),
         passwordChangeLinkAttributes: this._initiatePasswordChange()
+      },
+      uid: message.uid
+    })
+  }
+
+  Mailer.prototype.verifyLoginEmail = function (message) {
+    log.trace({ op: 'mailer.verifyLoginEmail', email: message.email, uid: message.uid })
+
+    var query = {
+      code: message.code,
+      uid: message.uid
+    }
+
+    if (message.service) { query.service = message.service }
+    if (message.redirectTo) { query.redirectTo = message.redirectTo }
+    if (message.resume) { query.resume = message.resume }
+
+    var link = this.verifyLoginUrl + '?' + qs.stringify(query)
+    query.one_click = true
+    var oneClickLink = this.verifyLoginUrl + '?' + qs.stringify(query)
+
+    return this.send({
+      acceptLanguage: message.acceptLanguage,
+      email: message.email,
+      headers: {
+        'X-Link': link,
+        'X-Service-ID': message.service,
+        'X-Uid': message.uid,
+        'X-Verify-Code': message.code
+      },
+      subject: gettext('Confirm new sign-in to Firefox'),
+      template: 'verifyLoginEmail',
+      templateValues: {
+        email: message.email,
+        device: this._formatUserAgentInfo(message),
+        link: link,
+        oneClickLink: oneClickLink,
+        supportUrl: this.supportUrl,
+        supportLinkAttributes: linkAttributes(this.supportUrl)
       },
       uid: message.uid
     })
--- a/partials/verify_login.html
+++ b/partials/verify_login.html
@@ -1,0 +1,42 @@
+<% extends "partials/base/base.html" %>
+
+<% block content %>
+<!--Header Area-->
+<tr style="page-break-before: always">
+  <td valign="top">
+    <h1 style="font-family: sans-serif; font-weight: normal; margin: 0 0 24px 0; text-align: center;"><% block verify_title %>{{t "New sign-in to Firefox"}}<% endblock %></h1>
+    <p class="primary" style="font-family: sans-serif; font-size: 14px; font-weight: normal; margin: 0 0 24px 0; text-align: center;"><% block verify_content %>{{t "For added security, please confirm this sign-in to begin syncing with %(device)s." }}<% endblock %></p>
+  </td>
+</tr>
+
+<!--Button Area-->
+<tr height="50">
+  <td align="center" valign="top">
+    <table border="0" cellpadding="0" cellspacing="0" height="100%" width="100%" id="email-button" style="-webkit-text-size-adjust: 100%; border-collapse: collapse; mso-table-lspace: 0pt; mso-table-rspace: 0pt; background-color: #0095dd; border-radius: 4px; height: 50px; width: 310px !important;">
+      <tr style="page-break-before: always">
+        <td align="center" valign="middle" id="button-content" style="font-family: sans-serif; font-weight: normal; text-align: center; margin: 0; color: #ffffff; font-size: 20px; line-height: 100%;">
+          <!--[if mso]>
+          <v:roundrect xmlns:v="urn:schemas-microsoft-com:vml" xmlns:w="urn:schemas-microsoft-com:office:word" href="{{{link}}}" style="width:280px;height:40px;v-text-anchor:middle;" arcsize="10%" stroke="f" fillcolor="#0095DD">
+            <w:anchorlock/>
+            <center>
+          <![endif]-->
+          <a href="{{{link}}}" id="button-link" style="font-family:sans-serif; color: #fff; display: block; padding: 15px; text-decoration: none; width: 280px;">{{t "Confirm sign-in"}}</a>
+          <!--[if mso]>
+          </center>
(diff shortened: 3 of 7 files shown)


partials/re_verify.html
```diff
@@ -0,0 +1,51 @@
+<% extends "partials/base/base.html" %>
+
+<% block content %>
+<!--Header Area-->
+<tr style="page-break-before: always">
+  <td valign="top">
+    <p class="primary" style="font-family: sans-serif; font-size: 14px; font-weight: normal; margin: 0 0 24px 0; text-align: center;"><% block verify_content %>{{{t "You signed in to Firefox from [OS], [Location, Location, Location] Please confirm this sign-in to begin syncing with [OS]." }}}<% endblock %></p>
```
shane-tomlinson: Can you replace `[OS]` with `%(deviceName)s` and `[Location, Locaiton, Location]` with `%(location)s`? We also need a period after `%(location)s`.

partials/re_verify.html
```diff
@@ -0,0 +1,51 @@
...
+  <td align="center" valign="top">
+    <table border="0" cellpadding="0" cellspacing="0" height="100%" width="100%" id="email-button" style="-webkit-text-size-adjust: 100%; border-collapse: collapse; mso-table-lspace: 0pt; mso-table-rspace: 0pt; background-color: #0095dd; border-radius: 4px; height: 50px; width: 310px !important;">
+      <tr style="page-break-before: always">
+        <td align="center" valign="middle" id="button-content" style="font-family: sans-serif; font-weight: normal; text-align: center; margin: 0; color: #ffffff; font-size: 20px; line-height: 100%;">
+          <!--[if mso]>
+          <v:roundrect xmlns:v="urn:schemas-microsoft-com:vml" xmlns:w="urn:schemas-microsoft-com:office:word" href="{{{link}}}" style="width:280px;height:40px;v-text-anchor:middle;" arcsize="10%" stroke="f" fillcolor="#0095DD">
+            <w:anchorlock/>
+            <center>
+          <![endif]-->
+          <a href="{{{link}}}" id="button-link" style="font-family:sans-serif; color: #fff; display: block; padding: 15px; text-decoration: none; width: 280px;">{{t "Confirm sign-in"}}</a>
```
shane-tomlinson: @ryanfeeley - in the content-server, we only use [sign in](https://github.com/mozilla/fxa-content-server/blob/a91d53d4b571f8d8c7a5b0e9d48cbbe1e5d7cb2f/app/scripts/templates/sign_in.mustache#L82), never hyphenated. Should we stay consistent here?

mailer.js
```diff
@@ -179,6 +181,45 @@ module.exports = function (log) {
     })
   }
 
+  Mailer.prototype.reVerifyEmail = function (message) {
```
rfk: Please call this something like "confirmLoginEmail" rather than using the term "reverify", which we found to cause a fair bit of confusion around this feature last week.

templates/re_verify.html
```diff
@@ -0,0 +1,67 @@
...
+<tr style="page-break-before: always">
+  <td align="center" id="firefox-logo" style="padding: 20px 0;">
+    <img src="http://image.e.mozilla.org/lib/fe9915707361037e75/m/2/fxlogojg.gif" height="95" width="88" alt="" style="-ms-interpolation-mode: bicubic;" />
+  </td>
+</tr>
+
+<!--Header Area-->
+<tr style="page-break-before: always">
+  <td valign="top">
+    <p class="primary" style="font-family: sans-serif; font-size: 14px; font-weight: normal; margin: 0 0 24px 0; text-align: center;">{{{t "You signed in to Firefox from [OS], [Location, Location, Location] Please confirm this sign-in to begin syncing with [OS]." }}}</p>
```
shane-tomlinson: Same comments about `OS` and `Location`, and the period here.

partials/re_verify.html
shane-tomlinson: edit: instead of `%(deviceName)s`, use `%(device)s` to be consistent with the suspicious location template.

There are no links here, no need for `{{{..}}}`. Please convert to `{{..}}`

templates/re_verify.html
shane-tomlinson: Though I suppose these are generated from the partials.

partials/re_verify.html
ryanfeeley: @shane-tomlinson Not the same word. "Sign-in" is to "sign in" as "login" is to "log in".

templates/re_verify.html
ryanfeeley: @shane-tomlinson There should be a period after {Location}, well spotted.

partials/re_verify.html
shane-tomlinson: @ryanfeeley - I don't follow.

shane-tomlinson: > @ryanfeeley - I don't follow.

Though to be fair, I've always been a little slow at English. And I definitely won't block on a hyphen.

ryanfeeley: @shane-tomlinson "Your login form would not let me log in." = "Your sign-in form would not let me sign in."

shane-tomlinson: > @shane-tomlinson "Your login form would not let me log in." = "Your sign-in form would not let me sign in."

I get it now! :+1:

partials/re_verify.html
vbudhram: @shane-tomlinson, @rfk This might be something that I missed, but not sure how we wanted to get location of user. Did we want show IP address and eventually do a location lookup?

shane-tomlinson: It'll be passed to the mailer, no need to worry about it here.

partials/confirm_login.html
```diff
@@ -0,0 +1,51 @@
+<% extends "partials/base/base.html" %>
+
+<% block content %>
+<!--Header Area-->
+<tr style="page-break-before: always">
+  <td valign="top">
+    <p class="primary" style="font-family: sans-serif; font-size: 14px; font-weight: normal; margin: 0 0 24px 0; text-align: center;"><% block verify_content %>{{t "You signed in to Firefox from %(device)s, %(location)s."}} {{t "Please confirm this sign-in to begin syncing with %(device)." }}<% endblock %></p>
```
shane-tomlinson: Thanks @vbudhram. I dig it. Now I'm going to be a jerk. I propose we add location in phase 2, not the initial roll-out, at least if we go for the end of the week release. My reasoning is that it'll add some amount of complexity, and if we go for the short dev timeline, we'll want as few moving parts as possible. Keeping two strings, one with location, one without, should provide us with sufficient flexibility w.r.t. l10n.

partials/confirm_login.html
```diff
@@ -0,0 +1,51 @@
...
+<% endblock %>
+
+<% block after_table %>
+<div itemscope itemtype="https://schema.org/EmailMessage">
+  <div itemprop="potentialAction" itemscope itemtype="https://schema.org/ViewAction">
+    <link itemprop="target" href="{{{oneClickLink}}}"/>
+    <meta itemprop="name" content="{{t 'Verify Email'}}"/>
+    <meta itemprop="url" content="{{{oneClickLink}}}"/>
+  </div>
+  <meta itemprop="description" content="{{t 'Verify your email to finish your Firefox Account registration'}}"/>
```
shane-tomlinson: Can you change `registration` to `email confirmation`?

shane-tomlinson: `email confirmation` is terrible. @ryanfeeley - should `registration` be `signin`, `sign in`, `sign-in`?

partials/confirm_login.html
shane-tomlinson: @vbudhram - I'm checking the text against @ryanfeeley's mocks. Can you replace `Please confirm this sign-in to begin syncing with %(device).` with `For added security, please confirm this sign-in to begin syncing with %(device).`

templates/confirm_login.txt
```diff
@@ -0,0 +1,4 @@
+{{t "You signed in to Firefox from %(device)s, %(location)s."}} {{t "Please confirm this sign-in to begin syncing with %(device)." }}
```
shane-tomlinson: `Please confirm` to `For added security, please confirm` here too.

partials/confirm_login.html
vbudhram: @shane-tomlinson I agree with you on adding location to phase 2 and separating out the location strings. Also, thanks for double checking string, must have overlooked this :(

partials/confirm_login.html
ryanfeeley: @shane-tomlinson @vbudhram I'm sure I presented all three necessary screens, but I lost track of where they are in these PRs. I didn't write that copy.

shane-tomlinson: @ryanfeeley - this is not for the email itself, this is for the future one click verification for Google.

config.js
```diff
@@ -95,6 +95,11 @@ var conf = convict({
       format: String,
       default: undefined
     },
+    verificationSigninUrl: {
```
shane-tomlinson: I have it on good authority (@philbooth) that the auth server endpoint to verify tokens is going to be called `/token/verify`. Can we call this `tokenVerifyUrl`?

mailer.js
```diff
@@ -179,6 +182,47 @@ module.exports = function (log) {
     })
   }
 
+  Mailer.prototype.verifyTokenEmail = function (message) {
+    log.trace({ op: 'mailer.verifyTokenEmail', email: message.email, uid: message.uid })
+
+    var query = {
+      code: message.code
+    }
```
shane-tomlinson: @philbooth - I forget, to verify token emails need the uid too?

shane-tomlinson:
Closing in favor of #1232 and #1233.

shane-tomlinson:
Agh, this was the wrong repo to delete the branch from.

mailer.js
shane-tomlinson: Phil told me "no, only the code"

shane-tomlinson:
@vbudhram - this has @ryanfeeley's latest text, I think this is in pretty good shape. What do you think?

config.js
```diff
@@ -156,11 +161,13 @@ var options = {
...
-conf.set('mail.initiatePasswordResetUrl', conf.get('contentServer.url') + '/reset_password')
-conf.set('mail.initiatePasswordChangeUrl', conf.get('contentServer.url') + '/settings/change_password')
+var contentServerUrl = conf.get('contentServer.url')
+conf.set('mail.signInUrl', contentServerUrl + '/signin')
+conf.set('mail.verificationUrl', contentServerUrl + '/v1/verify_email')
+conf.set('mail.passwordResetUrl', contentServerUrl + '/v1/complete_reset_password')
+conf.set('mail.accountUnlockUrl', contentServerUrl + '/v1/complete_unlock_account')
+conf.set('mail.initiatePasswordResetUrl', contentServerUrl + '/reset_password')
+conf.set('mail.initiatePasswordChangeUrl', contentServerUrl + '/settings/change_password')
+conf.set('mail.verifyLoginUrl', contentServerUrl + '/verify_signin')
```
shane-tomlinson: Let's use `complete_signin` for the route.

shane-tomlinson: Since we are using `signin` instead of `login` in the route, the config option should probably be `verifySigninUrl` too.

shane-tomlinson: Hmm, we have both login and signin. :/

ryanfeeley:
I think our stop-gap notification works better for at-a-glance usage.
h1. New sign-in to Firefox
**Firefox 48, Mac OS X 10.11**
**2016-04-13 20:01 UTC**
For added security, please confirm this sign-in to begin syncing.
button.Confirm sign-in

vbudhram:
Per meeting with @ryanfeeley, making slight updates to templates

subject: Confirm new sign-in to Firefox

h1: New sign-in to Firefox

For added security, please confirm this sign-in to begin syncing with {Firefox 45, Mac OS X 10.11}.

[ Confirm sign-in ]

Alternatively:
http://127.0.0.1:3030/verify_signin?code=7dacef4c38f44a446064c8b65ce715a2e93b6ab5ac1fbbdb50a161a8ab733c69&service=sync

vbudhram:
👍 Verified email is created as expected

screen shot 2016-04-18 at 4 28 13 pm

philbooth:
If this already got discussed, no worries, please ignore. But I'm just wondering whether that last footnote paragraph could be made more prominent. I still worry about people not reading the main bit properly and clicking the link without thinking is all.

vbudhram:
No, we didn't discuss that, @ryanfeeley thoughts?

vbudhram:
@shane-tomlinson @ryanfeeley Updated template footer per meeting, here is how it looks now
screen shot 2016-04-25 at 2 12 42 pm

Any objections to merging this in?

ryanfeeley:
👍

config.js
```diff
@@ -156,11 +161,13 @@ var options = {
...
-conf.set('mail.initiatePasswordResetUrl', conf.get('contentServer.url') + '/reset_password')
-conf.set('mail.initiatePasswordChangeUrl', conf.get('contentServer.url') + '/settings/change_password')
+var contentServerUrl = conf.get('contentServer.url')
+conf.set('mail.signInUrl', contentServerUrl + '/signin')
+conf.set('mail.verificationUrl', contentServerUrl + '/v1/verify_email')
+conf.set('mail.passwordResetUrl', contentServerUrl + '/v1/complete_reset_password')
+conf.set('mail.accountUnlockUrl', contentServerUrl + '/v1/complete_unlock_account')
+conf.set('mail.initiatePasswordResetUrl', contentServerUrl + '/reset_password')
+conf.set('mail.initiatePasswordChangeUrl', contentServerUrl + '/settings/change_password')
+conf.set('mail.verifyLoginUrl', contentServerUrl + '/complete_signin')
```
vladikoff: @shane-tomlinson @vbudhram do we need another patch to include the same in `fxa-auth-server`?

vbudhram: Nope, don't think so, @shane-tomlinson opened one a little while back. https://github.com/mozilla/fxa-auth-server/pull/1233, which I should be able to merge in shortly.

PR #143, opened 2016-05-02
fix(l10n): adjust l10n-extract keyword

vladikoff:
Fixes: https://github.com/mozilla/fxa-content-server/issues/3612

Ref: https://github.com/zaach/jsxgettext/pull/106


vladikoff:
Checked diff before / after extraction:

Issue #155, opened 2016-05-11
"password changed successfully" HTML template asks the user to change their password if the user did not change their password


shane-tomlinson:
As @ryanfeeley points out, If the account password has been changed by a malicious actor, the user won't know the new password. They should be displayed the reset link instead.

PR #156, opened 2016-05-11
fix(mailer): Add change password link to "Verify login" email

shane-tomlinson:
Also remove `passwordChangeLinkAttributes` from all emails that do not
use them.

@vbudhram - r?

### DIFF
--- a/mailer.js
+++ b/mailer.js
@@ -66,8 +66,8 @@ module.exports = function (log) {
     return linkAttributes(this.supportUrl)
   }
 
-  Mailer.prototype._initiatePasswordChange = function () {
-    return linkAttributes(this.initiatePasswordChangeUrl)
+  Mailer.prototype._initiatePasswordChange = function (email) {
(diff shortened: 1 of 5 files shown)


mailer.js
```diff
@@ -311,11 +299,11 @@ module.exports = function (log) {
       subject: gettext('Your Firefox Account password has been changed'),
       template: 'passwordChangedEmail',
       templateValues: {
+        passwordChangeLinkAttributes: this._initiatePasswordChange(message.email),
```
vbudhram: If a user's account password was changed, and it was not authorized, then they wouldn't know the new password to change it. The change password screen requires you to enter the current password and new password.

Would it make more sense to reset the password? This might be out of scope for this PR.

vbudhram:
LGTM 👍 , just the one minor question.

mailer.js
shane-tomlinson: That's a follow on PR, @ryanfeeley pointed out the same thing, so I opened #155.

PR #158, opened 2016-05-12
fix(mailer): Several mailer fixes

shane-tomlinson:
* Password changed template shows "reset your password" link. "change your password" used to be displayed. If an attacker changed the user's password, the user would not know the new password to change it.
* Fix the "reset password" link color in the "Your password has been reset" email.
* All password reset links auto-submit the password reset form.
* The "Change your password" link in the "New sign in to Firefox" email contains the users email address.

fixes #151
fixes #153
fixes #154
fixes #155
fixes #157

PR #161, opened 2016-05-20
feat(mailer): Add the `write-to-disk.js` script.

shane-tomlinson:
@philbooth - this is for you!

### DIFF
--- a/scripts/write-to-disk.js
+++ b/scripts/write-to-disk.js
@@ -32,14 +32,12 @@ var config = require('../config')
 var createMailer = require('../index')
 var fs = require('fs')
 var log = require('../log')('server')
+var mkdirp = require('mkdirp')
 var path = require('path')
 
 var OUTPUT_DIRECTORY = path.join(__dirname, '..', '.mail_output')
 
-function getEmailOutputPath(subject, extension) {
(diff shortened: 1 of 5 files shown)


scripts/write-to-disk.js
```diff
@@ -0,0 +1,121 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+/**
+ * Write emails to disk. Output is written to ./.mail_output/<email type>.html
```
philbooth: Maybe add a note that `./.mail_output` must exist in advance, it won't be created automatically? Or not, it becomes pretty obvious when you try to run the the script.

philbooth:
Sorry, I thought "this if for you" meant "this is for you to review #158 with" rather than "this is for you to review in addition to #158". 😊 

Anyway, works perfectly and looks good, r+.

scripts/write-to-disk.js
shane-tomlinson: I just added `mkdirp` as a `devDependency` to fix this. Thanks for pointing that out.

PR #163, opened 2016-05-30
feat(metrics): track templates used in sent email

### LINKED ISSUE #130: Add ability to track # of email sent
We need a way to tell how many emails were sent by `mailer.js`

vladikoff:
Fixes #130

### DIFF
--- a/mailer.js
+++ b/mailer.js
@@ -127,7 +127,11 @@ module.exports = function (log) {
       }, message.headers)
     }
 
-    log.info({ op: 'mailer.send', email: message.email })
+    log.info({
+      email: message.email,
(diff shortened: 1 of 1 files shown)


vladikoff:
@rfk r? this should be enough to get an overview of sent email into Kibana

vladikoff:
We would remove the `email` field from the dashboard and just use the template value.

vladikoff:
https://bugzilla.mozilla.org/show_bug.cgi?id=1276709

rfk:
LGTM, r+

> We would remove the email field from the dashboard and just use the template value.

IIUC this should happen automagically thanks to some PII filtering in heka.

PR #169, opened 2016-06-21
feat(metrics): include type of reminder in the email query

vladikoff:
@rfk r?

This will allow us to track verification flow via reminders on the auth-server

ref https://github.com/mozilla/fxa/issues/138


vladikoff:
image

PR #170, opened 2016-06-22
Grunt release scripting

rfk:
Per @jrgm's request at [1], this adds our usual grunt scripting for tagging releases, making it easier to deploy this as a standalone server instead of a library.  @vladikoff r?

[1] https://bugzilla.mozilla.org/show_bug.cgi?id=1281355#c2


vladikoff:
seems to work well, thanks!

PR #172, opened 2016-06-24
Fix eslint nags

jrgm:
There error is handled by defaulting to no output. So unlikely to happen. 

Merging since this fixes master, which is failing right now.


jrgm:
Crap. Two things on the branch. Closing.

PR #174, opened 2016-06-24
fix(bind): should by default bind to localhost

jrgm:
The default binding should be localhost; not a public endpoint, right? 

r? - @vladikoff

### DIFF
--- a/bin/server.js
+++ b/bin/server.js
@@ -83,8 +83,9 @@ P.all(
 
     api.listen(
       config.get('port'),
+      config.get('host'),
       function () {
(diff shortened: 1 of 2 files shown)


vladikoff:
Yeap, let me know if we need to fix fxa-dev

jrgm:
This should implicitly change fxa-dev to bind to localhost, right?

There's no need to have it listen on 0.0.0.0 in fxa-dev or fxa-local-dev, yes? 

(Yeah, there's firewalls and SGs, but don't listen on 0.0.0.0 unless you need to).

vladikoff:
👍

PR #176, opened 2016-06-24
chore(eslint): make eslint rules consistent

### LINKED ISSUE #175: Make eslint ignores consistent
Currently, we have `// eslint` in one file and `//eslint` in another file (both under `scripts/`). I suggest we remove these files from eslint.js so that we do not have to worry about consistency.
(Or lets make them consistent, that works as well)

@pdehaan

TDA:
Fixes #175

@pdehaan

### DIFF
--- a/scripts/.eslintrc
+++ b/scripts/.eslintrc
@@ -1,0 +1,6 @@
+{
+  "rules": {
+    "handle-callback-err": 0,
+    "no-console": 0
+  }
+}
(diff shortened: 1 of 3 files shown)


jrgm:
Hrm. I specifically chose to exempt `handle-callback-err` in `./scripts/rpm-version.js` for a specific case where I knew it didn't matter. 

Now you make that the rule for anything (including the future) under `./scripts`?

PR #179, opened 2016-07-04
feat(server): add db connection retry on start

### LINKED ISSUE #178: startup race between auth-db and auth-mailer
I was looking at the code, and it occurred to me that that there is a startup race between auth-db and auth-mailer. If auth-mailer starts and tries to connect before auth-db is accepting connections, auth-mailer will fail and never retry connect.

vladikoff:
Fixes #178

### DIFF
--- a/bin/server.js
+++ b/bin/server.js
@@ -10,13 +10,14 @@ var mailConfig = config.get('mail')
 
 var packageJson = require('../package.json')
 var P = require('bluebird')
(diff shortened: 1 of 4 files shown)


vladikoff:
@rfk @jrgm r?

rfk:
LGTM from a code POV, @jrgm is this the behaviour you had in mind?

Issue #162, opened 2016-05-24
Add location information to confirm signin emails


shane-tomlinson:
Follow on work. The original plans called for signin location information in the confirmation email. This was removed from the initial iteration so that we could move more quickly.

rfk:
Ref https://github.com/mozilla/fxa-auth-mailer/issues/165

rfk:
Moving this to the FxA-99 milestone

shane-tomlinson:
This can be closed since #180 has merged.

PR #180, opened 2016-07-08
feat(location): add location data to emails

TDA:
Adds the location data to emails
@vbudhram 
@vladikoff 
@shane-tomlinson

(@rfk adds: Connects to #162; this makes waffle link the PR with the issue card)


mailer.js
```diff
@@ -26,6 +27,28 @@ module.exports = function (log) {
...
+      if (location.city) {
+        locationString += location.city + ', '
+      }
+      locationString += location.country + ' ' + '(estimated)'
+    }
+    return locationString
+  }
+
+  function constructLocalTimeString (timeZone) {
+    // if no timeZone is passed, use 'America/Los_Angeles'
```
vbudhram: I don't believe we discussed this scenario in meeting. @ryanfeeley I feel like the fallback should be UTC since that is the time displayed now without locale information. Thoughts?

mailer.js
```diff
@@ -26,6 +27,28 @@ module.exports = function (log) {
...
+    moment.tz.setDefault('America/Los_Angeles');
```
vbudhram: Extra semicolon

vbudhram:
I would add some test cases similar to these [mailer tests](https://github.com/mozilla/fxa-auth-mailer/blob/master/test/local/mailer_tests.js#L274) to ensure that the formatting of the location and time are correct based on the different inputs.

vbudhram:
I believe the location data was also to be added to  `Sign-in confirmation` email templates. Unless you wanted to do another PR for it :)

mailer.js
```diff
@@ -26,6 +27,28 @@ module.exports = function (log) {
...
   }
 
+  function constructLocationString (location) {
+    // construct the location string from the location object
+    var locationString = ''
+    if (location !== null) {
+      if (location.city) {
+        locationString += location.city + ', '
+      }
+      locationString += location.country + ' ' + '(estimated)'
```
vladikoff: `(estimated)` string will need to be translated (or removed to keep thing simple)

templates/new_device_login.html
```diff
@@ -20,7 +20,13 @@
...
   <td valign="top">
     <h1 style="font-family: sans-serif; font-weight: normal; margin: 0 0 24px 0; text-align: center;">{{t "New sign-in to Firefox"}}</h1>
-    <p class="primary" style="font-family: sans-serif; font-size: 14px; font-weight: normal; margin: 0 0 12px 0; text-align: center;">{{{ device }}}<br/>{{ timestamp }}</p>
+    <p class="primary" style="font-family: sans-serif; font-size: 14px; font-weight: normal; margin: 0 0 12px 0; text-align: center;">{{{ device }}}<br/>
+
+<span style="display:block;">IP address: {{ ip }}</span>
+{{ location }}
+<span style="display:block;">{{ timestamp }}</span>
+
+</p>
```
vladikoff: a lot of extra space and line breaks added here

templates/new_device_login.txt
```diff
@@ -1,6 +1,8 @@
 {{t "New sign-in to Firefox" }}
 
 {{ device }}
+{{ ip }}
+{{ location }}
```
vladikoff: this might need some extra text

mailer.js
TDA: Would it be all right if I added {{t location}} in the templates? or should I call t() here? @vladikoff

templates/new_device_login.txt
TDA: Extra text as in? @vladikoff

mailer.js
vladikoff: https://github.com/mozilla/fxa-auth-mailer/search?utf8=%E2%9C%93&q=gettext

I think it should be `gettext` or something, check others

TDA:
@vbudhram Absolutely adding it to `sign-in confirmation` once we settle on the changes, then I will replicate it to the other pages as well :)

templates/new_device_login.txt
vladikoff: sorry, something like 
```
Device: {{ device }}
IP address: {{ ip }} 
```
?

vladikoff: cc @ryanfeeley

TDA: @vladikoff Its already there in the `.html` partials for IP, we arent adding any extra device info here, so I guess that stays that way?

vladikoff: up to @ryanfeeley, we need to test how the text email looks like. It might not be readable after we add more information to it.

mailer.js
TDA: Will Change :)

mailer.js
TDA: @vladikoff we discussed this in the mtg with @vbudhram and @ryanfeeley. `Estimated` is what we want to try out as the language for `not 100% accurate`. Will add the entire string to gettext() 👍

templates/new_device_login.txt
TDA: Understood! Up to the UX guru @rfeeley it is 💃

TDA:
@vbudhram Added tests, I hope thats what you had in mind. :)

templates/new_device_login.txt
rfk: A historical note: we created the original version of this email in a real hurry, and one of the constraints we worked under was that it not introduce any new strings.  Hence the raw undecorated device data being just plopped in the middle of the email like that.

We're no longer under that constraint, so perhaps @ryanfeeley can suggest something better.

test/local/mailer_tests.js
```diff
@@ -267,6 +267,34 @@ P.all(
...
+            ip: '219.129.234.194',
+            location: location,
+            timeZone: 'America/Los_Angeles'
+          }
+
+          test(
+            'Location and ip data is present in email template output for ' + type,
+            function (t) {
+              mailer.mailer.sendMail = function (emailConfig) {
+                t.ok(includes(emailConfig.html, location.city + ', ' + location.country))
```
vbudhram: Mind adding the location check to the `verifyLoginEmail` test block as well, https://github.com/mozilla/fxa-auth-mailer/blob/3c50620251260592a44e475649b2c76a9f9a7273/test/local/mailer_tests.js#L187

vbudhram: Actually, I think it makes more sense to pull these out and iterate over them similar to [this](https://github.com/mozilla/fxa-auth-mailer/blob/3c50620251260592a44e475649b2c76a9f9a7273/test/local/mailer_tests.js#L43) and [this](https://github.com/mozilla/fxa-auth-mailer/blob/3c50620251260592a44e475649b2c76a9f9a7273/test/local/mailer_tests.js#L95). That way if we add more templates that have location, we don't have to duplicate code.

TDA: @vbudhram Done 💃

mailer.js
```diff
@@ -96,6 +122,14 @@ module.exports = function (log) {
     return parts.join(', ')
   }
 
+  Mailer.prototype._constructLocationString = function (location) {
+    return constructLocationString(location)
+  }
+
+  Mailer.prototype._constructLocalTimeString = function (timestamp, locale) {
+    return constructLocalTimeString(timestamp, locale)
```
rfk: The function signature is `function constructLocalTimeString (timeZone, locale)` but here we're passing a first argument called `timestamp` rather than `timeZone`...is one of these mis-named?

partials/new_device_login.html
```diff
@@ -1,4 +1,8 @@
 <% extends "partials/base/base_text_only.html" %>
 
 <% block heading %>{{t "New sign-in to Firefox"}}<% endblock %>
-<% block text_primary %>{{{ device }}}<br/>{{ timestamp }}<% endblock %>
+<% block text_primary %>{{{ device }}}<br/>
+<span style="display:block;">IP address: {{ ip }}</span>
```
rfk: Yes, and me may also want to put the ":" in there for translation as well.  See for example `{{t "Alternatively:"}}` in other templates.

templates/verify_login.html
```diff
@@ -20,7 +20,12 @@
 <tr style="page-break-before: always">
   <td valign="top">
     <h1 style="font-family: sans-serif; font-weight: normal; margin: 0 0 24px 0; text-align: center;">{{t "New sign-in to Firefox"}}</h1>
-    <p class="primary" style="font-family: sans-serif; font-size: 14px; font-weight: normal; margin: 0 0 24px 0; text-align: center;">{{t "For added security, please confirm this sign-in to begin syncing with %(device)s." }}</p>
+    <p class="primary" style="font-family: sans-serif; font-size: 14px; font-weight: normal; margin: 0 0 24px 0; text-align: center;">{{t "For added security, please confirm this sign-in to begin syncing with this device:"}}<br/><br/>
+    {{{ device }}}
+    <span style="display:block;">IP address: {{ ip }}</span>
```
rfk: Likewise "IP Address" needs to be localizable here

templates/new_device_login.txt
```diff
@@ -1,6 +1,8 @@
 {{t "New sign-in to Firefox" }}
 
 {{ device }}
+{{ ip }}
```
rfk: The text-only variants should also say "IP Address: 1.2.3.4" for consistency.

mailer.js
```diff
@@ -26,6 +27,31 @@ module.exports = function (log) {
...
   }
 
+  function constructLocationString (location) {
+    // construct the location string from the location object
+    var locationString = ''
+    if (location) {
+      if (location.city) {
+        locationString += location.city + ', '
+      }
+      locationString += location.country + ' ' + '(estimated)'
```
rfk: `'(estimated)'` will also need to be localizable; in the first instance I'm going to assume we can just leave it off of the string here, and include it in the templates like this:

```
{{ location }} {{t (estimated) }}
```

mailer.js
rfk: Looks like it's called with `message.timeZone`, so the `timestamp` argument here is the odd one out.

mailer.js
```diff
@@ -219,12 +253,15 @@ module.exports = function (log) {
       templateValues: {
         device: this._formatUserAgentInfo(message),
         email: message.email,
+        ip: message.ip,
         link: link,
+        location: gettext(this._constructLocationString(message.location)),
```
rfk: Oh, I see this tries to localize the *whole* location string here.  I'm pretty sure this won't work out, as gettext can only be called with static(ish) strings rather than something built at runtime like this.  I think you'll need to split this up into a dynamic `location` string built at runtime, and a static `"(estimated)"` or similar that can be handled by the translation machinery.

mailer.js
TDA: @rfk Will i need conditionals in the templates to display this? Seeing as we dont display this if we dont have a location

rfk: Using a conditional sounds OK to me, I assume the same syntax as used in content-server views will also work here.

shane-tomlinson: > Using a conditional sounds OK to me, I assume the same syntax as used in content-server views will also work here.

From an l10n perspective, both the current and proposed schemes are verboten. Both make assumptions about the how the city, country, and `estimated` strings should be combined for all locales.

I propose we use two distinct strings and allow localizers to format the strings as they best see fit:

```js
if (! location) {
  return ''
}

if (location.city) {
  return t('%(city)s, %(country}s (estimated)', location);
} else {
  return t('%(country}s (estimated)', location);
}
```

`t` in this case would need to do the string interpolation.

I believe similar logic can be used within the templates - I have not tried this, so I have no idea:

```html
{{#location}}
  <span class="location">
  {{#location.city}}
     {{t "Location: %(city)s, %(country)s (estimated)", location}}
  {{/location.city}}
  {{^location.city}}
     {{t "Location: %(country)s (estimated)", location}}
  {{/location.city}}
  </span>
{{/location}}
```

It may be that we have to update the template's `t` helper function to do interpolation. It could also be that I don't really know mustache all that well.

Do not follow the [format used in the suspicious_location.html](https://github.com/mozilla/fxa-auth-mailer/blob/master/templates/suspicious_location.html#L39) template. The separate strings in `{{t "Location: }} {{ location }}` were used because those emails were originally written by the legal team and translated by a special legal translating service w/o our input on best practices for l10n. We haven't gone back and updated the templates (see https://github.com/mozilla/fxa-auth-mailer/issues/183).

mailer.js
```diff
@@ -26,6 +27,31 @@ module.exports = function (log) {
...
+        locationString += location.city + ', '
+      }
+      locationString += location.country + ' ' + '(estimated)'
+    }
+    return locationString
+  }
+
+  function constructLocalTimeString (timeZone, locale) {
+    // if no timeZone is passed, use 'Etc/GMT'
+    moment.tz.setDefault('Etc/GMT')
```
shane-tomlinson: Should it be GMT or UTC?

mailer.js
```diff
@@ -375,11 +406,13 @@ module.exports = function (log) {
       template: 'newDeviceLoginEmail',
       templateValues: {
         device: this._formatUserAgentInfo(message),
+        ip: message.ip,
+        location: gettext(this._constructLocationString(message.location)),
```
shane-tomlinson: See [this comment I made above](https://github.com/mozilla/fxa-auth-mailer/pull/180/files/3bc4a05af22f535f79e546e95b828802e1a3e7d9#r71302053) about translating the location.

partials/new_device_login.html
shane-tomlinson: Yees, I would format this like:

```
{{t "IP address: %(ip)s" }}
```

Again, we have to ensure the `t` helper function can do interpolation.

mailer.js
shane-tomlinson: This file already has a `gettext` function, but it doesn't do any translation, so we may have to call something like `this.translator.format(this.translator.gettext(<string to translate>), location)`  to actually do the translations.

partials/new_device_login.html
shane-tomlinson: I'm not sure how the [t helper function](https://github.com/mozilla/fxa-auth-mailer/blob/master/templates/index.js#L6) goes about its thing. You might have to [dig into the handlebars documentation](http://handlebarsjs.com/) to figure out whether t does what we need it to, and if not, how to make it do so.

mailer.js
TDA: @shane-tomlinson Whats the difference? :O IANA says both are equivalent :confused:

shane-tomlinson: https://en.wikipedia.org/wiki/Coordinated_Universal_Time and https://en.wikipedia.org/wiki/Greenwich_Mean_Time - UTC is generally considered the successor to GMT, and is what timezone offsets are calculated from. Until I read those articles, I always thought GMT observed daylight savings time, but it doesn't.

TDA:
@shane-tomlinson I *think* I fixed it, it *seems* to work, but I have no idea if this is what you wanted :O 

```
+  Mailer.prototype._constructLocationString = function (message) {
 +    var translator = this.translator(message.acceptLanguage)
 +    var location = message.location
 +    // construct the location string from the location object
 +    if (location) {
 +      if (location.city) {
 +        return translator.format(translator.gettext('%(city)s, %(country)s (estimated)'), location);
 +      } else {
 +        return translator.format(translator.gettext('%(country)s (estimated)'), location);
 +      }
 +    }
 +    return ''
 +  }
```

partials/new_device_login.html
```diff
@@ -1,4 +1,8 @@
 <% extends "partials/base/base_text_only.html" %>
 
 <% block heading %>{{t "New sign-in to Firefox"}}<% endblock %>
-<% block text_primary %>{{{ device }}}<br/>{{ timestamp }}<% endblock %>
+<% block text_primary %>{{{ device }}}<br/>
+<span style="display:block;">{{t "IP address: %(ip)s" }}</span>
+{{ location }}
```
shane-tomlinson: Is there an advantage to a `span` with `display:block` over a block level element like a `div` or `p`? Can you wrap location with a span/div as well, for consistency?

@ryanfeeley - Are location and timestamp supposed to have labels?

mailer.js
```diff
@@ -26,6 +27,19 @@ module.exports = function (log) {
     return 'href="' + url + '" style="color: #0095dd; text-decoration: none; font-family: sans-serif;"'
   }
 
+  function constructLocalTimeString (timeZone, locale) {
+    // if no timeZone is passed, use 'Etc/UTC'
+    moment.tz.setDefault('Etc/UTC')
+    // if no locale is passed, use 'en'
+    locale = locale || 'en'
+    moment.updateLocale(locale)
```
shane-tomlinson: [According to the moment docs, updateLocale](http://momentjs.com/docs/#/i18n/) is to `to update a locale's properties.` This does not change the locale used for translations. Instead `moment.locale` must be used. When I test with Spanish, updateLocale always prints the date in English. When I use `update`, Spanish is used.

## with updateLocale date is in English

screen shot 2016-07-22 at 11 17 14

## with update, it's in Spanish

screen shot 2016-07-22 at 11 17 47

mailer.js
```diff
@@ -26,6 +27,19 @@ module.exports = function (log) {
...
+    var time = moment()
+    if (timeZone) {
+      time = time.tz(timeZone)
+    }
+    // return a locale-specific time
+    return time.format('LTS (z) dddd, ll')
```
shane-tomlinson: This is only a pseudo-locale-specific time. We are specifying the format that says time must come first, then timezone, then full day of the week, then localized date.

We should consider using `LLLL` or `llll` which will translate all that information into the format best suited for that locale, with the caveat that the result will lack the user's timezone.


## LLLL

screen shot 2016-07-22 at 11 22 47

## ll

screen shot 2016-07-22 at 11 22 58

For v1, I'm not going to block on this.

mailer.js
```diff
@@ -219,12 +251,15 @@ module.exports = function (log) {
...
+        ip: message.ip,
         link: link,
+        location: this._constructLocationString(message),
         oneClickLink: oneClickLink,
         passwordChangeLink: this.createPasswordChangeLink(message.email),
         passwordChangeLinkAttributes: this._passwordChangeLinkAttributes(message.email),
         supportLinkAttributes: linkAttributes(this.supportUrl),
-        supportUrl: this.supportUrl
+        supportUrl: this.supportUrl,
+        timestamp: this._constructLocalTimeString(message.timeZone, message.acceptLanguage)
```
shane-tomlinson: Passing `message.acceptLanguage` does not work because `acceptLanguage` is a header that can contain several locales, e.g., `es;q=0.8,en-US;q=0.5,en;q=0.3`. When I tested with Spanish as my primary language and English as my secondary, this is the output I see is in English:

screen shot 2016-07-22 at 10 58 54

Instead, what I think we should do is construct a translator instance, pass the acceptLanguage to the translator, get `translator.language` which will be the best match language, best match for our own translations, and then pass that as the locale to moment. 

I did that by applying this patch:

```diff
diff --git a/mailer.js b/mailer.js
index 652927d..74a4a3d 100644
--- a/mailer.js
+++ b/mailer.js
@@ -124,12 +125,14 @@ module.exports = function (log) {
     return ''
   }
 
-  Mailer.prototype._constructLocalTimeString = function (timeZone, locale) {
-    return constructLocalTimeString(timeZone, locale)
+  Mailer.prototype._constructLocalTimeString = function (timeZone, acceptLanguage) {
+    var translator = this.translator(acceptLanguage)
+    return constructLocalTimeString(timeZone, translator.language)
   }
```

With that, the results are correctly translated.

shane-tomlinson:
There is a script called [scripts/write-to-disk.js](https://github.com/mozilla/fxa-auth-mailer/blob/master/scripts/write-to-disk.js) that can be used to write emails to disk. I have modified it slightly to test l10n:

```diff
diff --git a/scripts/write-to-disk.js b/scripts/write-to-disk.js
index 9c694bf..5245ca2 100644
--- a/scripts/write-to-disk.js
+++ b/scripts/write-to-disk.js
@@ -82,14 +81,20 @@ function sendMail(mailer, messageToSend) {
   var messageSubType = parts[1]
 
   var message = {
-    acceptLanguage: 'en',
+    acceptLanguage: 'es;q=0.8,en-US;q=0.5,en;q=0.3"',
     code: 'ae35999f861ffc81d594034eb4560af8',
     email: 'testuser@testuser.com',
+    ip: '10.246.67.38',
+    location: {
+      city: 'Madrid',
+      country: 'Spain'
+    },
     locations: [],
     redirectTo: 'https://redirect.com/',
     resume: 'eyJjYW1wYWlnbiI6bnVsbCwiZW50cnlwb2ludCI6bnVsbCwiZmxvd0lkIjoiM2Q1ODZiNzY4Mzc2NGJhOWFiNzhkMzMxMTdjZDU4Y2RmYjk3Mzk5MWU5NTk0NjgxODBlMDUyMmY2MThhNmEyMSIsInJlc2V0UGFzc3dvcmRDb25maXJtIjp0cnVlLCJ1bmlxdWVVc2VySWQiOiI1ODNkOGFlYS00NzU3LTRiZTQtYWJlNC0wZWQ2NWZhY2Y2YWQiLCJ1dG1DYW1wYWlnbiI6bnVsbCwidXRtQ29udGVudCI6bnVsbCwidXRtTWVkaXVtIjpudWxsLCJ1dG1Tb3VyY2UiOm51bGwsInV0bVRlcm0iOm51bGx9',
     service: 'sync',
     token: '47b22cd271963448cf36da95cccfcfb342b5693d66f58aa635f9a95579431002',
+    timeZone: 'Europe/Madrid',
     type: messageSubType,
     uaBrowser: 'Firefox',
     uaBrowserVersion: '47',
```

You can run the tool using:

```bash
➜  fxa-auth-mailer git:(feat-add-location-data) ✗ node ./scripts/write-to-disk.js verifyLoginEmail
```
The output will be in the .mail_output subdirectory.

To test Spanish, I had to add `es` to the list of [supported locales in config.js](https://github.com/mozilla/fxa-auth-mailer/blob/a63ca05821fb03bc0209fd9ca4ce235a2ebfcd12/config.js#L77).

The location strings look good, and they extract correctly to the l10n repo.

I find the 4 lines with the new information a bit too closely spaced, perhaps add a pixel or two between them:

screen shot 2016-07-22 at 11 22 58

mailer.js
```diff
@@ -26,6 +27,19 @@ module.exports = function (log) {
     return 'href="' + url + '" style="color: #0095dd; text-decoration: none; font-family: sans-serif;"'
   }
 
+  function constructLocalTimeString (timeZone, locale) {
+    // if no timeZone is passed, use 'Etc/UTC'
+    moment.tz.setDefault('Etc/UTC')
+    // if no locale is passed, use 'en'
+    locale = locale || 'en'
```
vladikoff: move `en` above above `module.exports`. (`DEFAULT_LOCALE`?), update the comment

mailer.js
```diff
@@ -26,6 +27,19 @@ module.exports = function (log) {
     return 'href="' + url + '" style="color: #0095dd; text-decoration: none; font-family: sans-serif;"'
   }
 
+  function constructLocalTimeString (timeZone, locale) {
+    // if no timeZone is passed, use 'Etc/UTC'
+    moment.tz.setDefault('Etc/UTC')
```
vladikoff: same as https://github.com/mozilla/fxa-auth-mailer/pull/180/files#r71884957

package.json
```diff
@@ -48,9 +48,10 @@
...
     "mkdirp": "0.5.1",
-    "proxyquire": "1.7.10",
-    "simplesmtp": "0.3.35",
-    "sinon": "1.17.4",
-    "tap": "6.2.0"
+    "moment-timezone": "0.5.4",
+    "proxyquire": "1.7.4",
+    "simplesmtp": "0.3.32",
+    "sinon": "1.17.3",
+    "tap": "5.4.2"
```
vladikoff: tap being downgraded?

vladikoff: as well as other dependencies ?

TDA: I blame my bad rebasing skills

partials/new_device_login.html
TDA: @shane-tomlinson the div/p will result in different styles, which I will need to override again. Span will not cause that. Also, the reason for location to not be wrapped is to save the conditional check for existence of location, as otherwise an empty new line will be presented.

mailer.js
TDA: @shane-tomlinson From same doc, http://momentjs.com/docs/#/customization/ , it says `2.12.0 deprecated using moment.locale() to change an existing locale. Use moment.updateLocale() instead.`
Should I go ahead and use `moment.locale`?

mailer.js
TDA: > We are specifying the format that says time must come first, then timezone, then full day of the week, then localized date.

IIRC, I believe this is exactly what @ryanfeeley asked for in the meeting about localization.

mailer.js
TDA: Going ahead with moment.locale() for now.

TDA:
@shane-tomlinson and @vladikoff 
Changes made:
1. Added `en` and `utc` as defaults 
2. Added a margin between the lines in the email
3. Changed `updateLocale` to `locale`
4. Use a translator instance to get the language
5. Added conditional, with span block around `location` to be consistent with the other lines

Have tested locally to see that they work as expected :)

partials/base/base.html
```diff
@@ -3,6 +3,12 @@
 <head>
   <meta http-equiv="Content-Type" content="text/html; charset=UTF-8" />
   <title>{{t "Firefox Accounts"}}</title>
+  <style>
+    .primary > span {
+      /* only affects the span's that are set to display:block; */
+      margin: 2px auto;
```
shane-tomlinson: We can't depend on styles working in a lot of mail clients, instead the styles need to be added to each span. Also a good reason to [surround the location with a span](https://github.com/mozilla/fxa-auth-mailer/pull/180#discussion_r71849114).

templates/new_device_login.html
```diff
@@ -3,6 +3,12 @@
 <head>
   <meta http-equiv="Content-Type" content="text/html; charset=UTF-8" />
   <title>{{t "Firefox Accounts"}}</title>
+  <style>
+    .primary > span {
+      /* only affects the span's that are set to display:block; */
+      margin: 2px auto;
```
shane-tomlinson: Same not here - We can't depend on styles working in a lot of mail clients, instead the styles need to be added to each span.

templates/new_device_login.html
```diff
@@ -20,7 +26,15 @@
 <tr style="page-break-before: always">
   <td valign="top">
     <h1 style="font-family: sans-serif; font-weight: normal; margin: 0 0 24px 0; text-align: center;">{{t "New sign-in to Firefox"}}</h1>
-    <p class="primary" style="font-family: sans-serif; font-size: 14px; font-weight: normal; margin: 0 0 12px 0; text-align: center;">{{{ device }}}<br/>{{ timestamp }}</p>
+    <p class="primary" style="font-family: sans-serif; font-size: 14px; font-weight: normal; margin: 0 0 12px 0; text-align: center;">{{{ device }}}<br/>
+<span style="display:block;">{{t "IP address: %(ip)s" }}</span>
+{{#if location }}
```
shane-tomlinson: Ahha, you've done it.

templates/new_device_login.txt
```diff
@@ -1,6 +1,8 @@
 {{t "New sign-in to Firefox" }}
 
 {{ device }}
+{{t "IP address: %(ip)s" }}
+{{ location }}
```
shane-tomlinson: Does this leave an empty line if location is not available?

templates/verify_login.txt
```diff
@@ -1,5 +1,9 @@
 {{t "New sign-in to Firefox"}}
-{{t "For added security, please confirm this sign-in to begin syncing with %(device)s." }}
+{{t "For added security, please confirm this sign-in to begin syncing with this device:" }}
+{{ device }}
+{{t "IP address: %(ip)s" }}
+{{ location }}
```
shane-tomlinson: Will this leave an empty line if location is not available?

shane-tomlinson: It does:

screen shot 2016-07-25 at 15 37 01

templates/new_device_login.txt
shane-tomlinson: Even with an `if` around this line, the extra line appears if there is no location information. Not sure what the solution is.

templates/verify_login.txt
shane-tomlinson: Even with an if around this line, the extra line appears if there is no location information. Not sure what the solution is.

Let's leave this for now.

shane-tomlinson:
@TDA - aside from the use of the `style` element, this looks in pretty good shape. I'm going to make that change now so we can get this merged. Thanks for the several rounds of iteration, great work!

shane-tomlinson:
Email on acid test here: https://www.emailonacid.com/app/acidtest/display/summary/FfgBey79quzNrVWv1aufaMu5OJg0K572f5UIrKUMHzqlk

shane-tomlinson:
In Outlook < 2016, each of the info lines is on the same line.  This may be due to block level elements being inside of a `p`, `p` is not allowed to contain other block level elements. File a follow up for this.

partials/verify_login.html
```diff
@@ -5,7 +5,12 @@
 <tr style="page-break-before: always">
   <td valign="top">
     <h1 style="font-family: sans-serif; font-weight: normal; margin: 0 0 24px 0; text-align: center;"><% block verify_title %>{{t "New sign-in to Firefox"}}<% endblock %></h1>
-    <p class="primary" style="font-family: sans-serif; font-size: 14px; font-weight: normal; margin: 0 0 24px 0; text-align: center;"><% block verify_content %>{{t "For added security, please confirm this sign-in to begin syncing with %(device)s." }}<% endblock %></p>
+    <p class="primary" style="font-family: sans-serif; font-size: 14px; font-weight: normal; margin: 0 0 24px 0; text-align: center;"><% block verify_content %>{{t "For added security, please confirm this sign-in to begin syncing with this device:"}}<br/><br/>
+    {{{ device }}}
+    <span style="display:block;margin:2px 0;">{{t "IP address: %(ip)s" }}</span>
```
shane-tomlinson: @TDA - I know this is late in the process - to prepare for phased rollout, can you wrap both the IP address and timestamp lines in `{{if #`'s similar to how `location` is done, and repeat everywhere this info has been added? Can you do that for both the html and text variants of verify_login and new_device_login?

TDA:
@shane-tomlinson Its done :)

package.json
```diff
@@ -48,6 +48,7 @@
     "jsxgettext-recursive": "0.0.5",
     "load-grunt-tasks": "3.5.0",
     "mkdirp": "0.5.1",
+    "moment-timezone": "0.5.4",
```
shane-tomlinson: This looks to me like a hard dependency instead of a devDependency. Can you move this to `dependencies` and re-run shrinkwrap?

partials/new_device_login.html
```diff
@@ -1,4 +1,8 @@
 <% extends "partials/base/base_text_only.html" %>
 
 <% block heading %>{{t "New sign-in to Firefox"}}<% endblock %>
-<% block text_primary %>{{{ device }}}<br/>{{ timestamp }}<% endblock %>
+<% block text_primary %>{{{ device }}}<br/>
+{{#if ip }}<span style="display:block;margin:2px 0;">{{t "IP address: %(ip)s" }}</span>{{/if}}
```
shane-tomlinson: 👍 Thanks for wrapping.

TDA:
@shane-tomlinson Moved dep and updated shrinkwrap :)

shane-tomlinson:
🎉 Well done @TDA, and thanks for the reviews @vladikoff, @vbudhram and @rfk.

rfk:
I concur: 🎉

PR #181, opened 2016-07-14
fix(mailer): Slow down the verification reminder emails in dev.

shane-tomlinson:
This fixes the content-server functional tests for unverified accounts. The
verification reminder emails arrived before sign in w/ an unverified account
verification emails.

fixes mozilla/fxa-content-server#3929

@vladikoff - r? ~~Can you hold off on merging, I want to make sure this really does what it says it does.~~

### DIFF
--- a/config/dev.json
+++ b/config/dev.json
@@ -4,8 +4,8 @@
     "fmt": "pretty"
   },
   "verificationReminders": {
-    "reminderTimeFirst": 1000,
-    "reminderTimeSecond": 5000,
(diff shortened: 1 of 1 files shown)


shane-tomlinson:
@vladikoff - I set the timeout to 10 and 15 seconds instead of 5 and 10. Some functional tests sign up a user, then wait 5 seconds to ensure no WebChannel messages have arrived, then try to sign in. With a 5 second delay, sometimes the first verification reminder arrived before the newest signup verification email.

PR #187, opened 2016-07-21
feat(mailer): Remove the account lockout feature.

### LINKED ISSUE #186: Remove the account lockout feature
With signin confirmation and email caphta we have more targeted
meant to keep bad people at bay while minimizing friction for valid
users, account lockout no longer seems particularly elegant.

Ref https://github.com/mozilla/fxa-content-server/issues/3949

shane-tomlinson:
With signin confirmation and email recaphta we have more targeted
meant to keep bad people at bay while minimizing friction for valid
users, account lockout no longer seems particularly elegant.

fixes #186 

This should not be merged until after https://github.com/mozilla/fxa-auth-server/pull/1360


vladikoff:
r+ on this one, merge after https://github.com/mozilla/fxa-auth-server/pull/1360 is done

PR #190, opened 2016-07-26
feat(metrics): Add utm_* metrics

### LINKED ISSUE #182: Include utm_* metrics params in all our email links
In [1] our success metrics include the ability to see whether users are clicking through links from our security-related emails.  Let's support this by including utm_* on all embedded links in our emails.

We have prior art for this in the post-verification email, which includes the following query params:

```
utm_source=email&utm_medium=email&utm_campaign=fx-account-verified
```

I like the pattern here of using `utm_source` and `utm_medium` to indicate a link of followed from an email, and `utm_campaign` to indicate which email.

[1] https://github.com/mozilla/fxa/tree/master/features/proposed/FxA-99-improved-login-notification#metrics

### LINKED ISSUE #188: Add a query parameter to the "alternatively" link to log if anybody actually uses it.
Another of @ryanfeeley's great ideas that is being typed up by me. The alternatively link is gigantic and ugly. If nobody (or very few people) are using it, we should get rid of it. How do we know if people are using it? We add a new query parameter to the alternatively link only, making the link longer. If the query parameter is there, then we know the user copied that link.

vbudhram:
This PR adds a bunch of UTM params to all our email links. [Here is a list](https://github.com/mozilla/fxa-auth-mailer/issues/182#issuecomment-234312264) of the params and the links on each email. 

Fixes #182 
FIxes #188


mailer.js
```diff
@@ -162,9 +169,11 @@ module.exports = function (log) {
     if (message.redirectTo) { query.redirectTo = message.redirectTo }
     if (message.resume) { query.resume = message.resume }
 
-    var link = this.verificationUrl + '?' + qs.stringify(query)
+    var link = this._generateUTMLink(this.verificationUrl, query, template, 'activate')
+    var alternativeLink = this._generateUTMLink(this.verificationUrl, query, template, 'activate-alternative')
+
     query.one_click = true
-    var oneClickLink = this.verificationUrl + '?' + qs.stringify(query)
+    var oneClickLink = this._generateUTMLink(this.verificationUrl, query, template, 'activate')
```
shane-tomlinson: Since this will eventually be a button that is displayed within GMail itself and we'll want to know the click rate of those separately from everything else, should `activate` be `oneclick-activate`?

mailer.js
```diff
@@ -61,8 +67,8 @@ module.exports = function (log) {
     this.mailer.close()
   }
 
-  Mailer.prototype._supportLinkAttributes = function () {
-    return linkAttributes(this.supportUrl)
+  Mailer.prototype._supportLinkAttributes = function (template) {
+    return linkAttributes(this._generateUTMLink(this.supportUrl, {}, template, 'support'))
```
shane-tomlinson: `support` here is vague and won't allow us to determine which email produced the click. Can `support` be tied to the template somehow, e.g., `activate-support` when called from `verifyEmail`?

mailer.js
shane-tomlinson: Instead of passing in `activate` and `activate-alternative` and `support`, it seems like you could [create another map, this time of template name to context](https://github.com/mozilla/fxa-auth-mailer/pull/190/files#diff-3427c5e3d78d8d6550178ad46b9e0680R11). The last parameter to _generateUTMLink could then be a context suffix instead of the context itself. A map could be used to [implement the support link issue mentioned above](https://github.com/mozilla/fxa-auth-mailer/pull/190/files#r72398196).

mailer.js
```diff
@@ -61,8 +67,8 @@ module.exports = function (log) {
     this.mailer.close()
   }
 
-  Mailer.prototype._supportLinkAttributes = function () {
-    return linkAttributes(this.supportUrl)
+  Mailer.prototype._supportLinkAttributes = function (template) {
```
shane-tomlinson: I found the variable name `template` in these methods confusing, at first I was unsure if `template` was the template name or the actual template. Could you rename the variable to `templateName` or similar, where it's used as a name, to help with clarity?

mailer.js
```diff
@@ -456,12 +472,32 @@ module.exports = function (log) {
...
+
+    queryParams['utm_source'] = 'email'
+    queryParams['utm_medium'] = 'email'
+
+    var campaign = utmCampaignMap[template]
+    if (campaign) {
+      queryParams['utm_campaign'] = utmPrefix + campaign
+    }
+
+    if (context) {
```
shane-tomlinson: For the idea above about fetching context from a map, it could be implemented here identically to how the campaign is fetched from `utmCampaignMap`.

```js
var context = utmContextMap[template]
if (context) {
  queryParams['utm_context'] = utmPrefix + context + contextSuffix
}
```

mailer.js
```diff
@@ -7,6 +7,12 @@ var P = require('bluebird')
 var nodemailer = require('nodemailer')
 
 module.exports = function (log) {
+  // Email template to UTM campaign map
+  var utmCampaignMap = {
```
shane-tomlinson: I wonder if the intent would be clearer if this were named `templateNameToCampaignMap` - it's verbose, but which item is the key and which is the value is clear.

shane-tomlinson:
@vbudhram - I know this PR is labelled WIP and I'm jumping the gun a bit - I'm reviewing early so I can add support link click reporting to the content server and want to make sure everything is in place to make it so!

shane-tomlinson:
> is in place to make it so!

Incomplete thought. With a couple little changes, everything will be in place. :D Thanks for jumping on this.

mailer.js
vbudhram: My original idea was that template would signify the email it came from and context would be what the user did in the email, [utm param breakdown](https://github.com/mozilla/fxa-auth-mailer/issues/182#issuecomment-234312264). For example, the support link in the account verified email would look something like this

`https://support.mozilla.org/kb/im-having-problems-with-my-firefox-account?utm_source=email&utm_medium=email&utm_campaign=fxa-welcome&utm_context=fxa-support`

That being said, there is probably a better naming convention for the values of the params.

vbudhram:
@shane-tomlinson Thanks for taking a look, made all updates you recommended, except mapping to the context map. I am still a little confused about how it would be helpful, possible to set aside a little time to clear up confusion tomorrow and go over?

Besides that, I updated all the remaining email templates. All links leaving emails have params that correspond to https://github.com/mozilla/fxa-auth-mailer/issues/182#issuecomment-234312264. My plan is to block this until https://github.com/mozilla/fxa-auth-mailer/pull/180 gets in, since there will be merge conflicts.

Would you mind taking another look? While unit tests are there, I still need to do some manually/acid testing on these.

mailer.js
shane-tomlinson: Ahha, I was confused, your use makes perfect sense. I want the utm parameter on the [passwordChangeLink](https://github.com/mozilla/fxa-auth-mailer/pull/190/files/22e7a0840540ca104090e636f3529c73c1ff8766#diff-3427c5e3d78d8d6550178ad46b9e0680R234) and [passwordResetLink](https://github.com/mozilla/fxa-auth-mailer/pull/190/files/22e7a0840540ca104090e636f3529c73c1ff8766#diff-3427c5e3d78d8d6550178ad46b9e0680R290).

mailer.js
```diff
@@ -450,34 +504,72 @@ module.exports = function (log) {
...
+    query['utm_medium'] = 'email'
+
+    var campaign = templateNameToCampaignMap[templateName]
+    if (campaign && !query['utm_campaign']) {
+      query['utm_campaign'] = utmPrefix + campaign
+    }
 
-    return this.initiatePasswordResetUrl + '?' + qs.stringify(queryParams)
+    if (context) {
+      query['utm_context'] = utmPrefix + context
```
vbudhram: Per meeting, update to `utm_content`, to better match ga params.

vbudhram:
@shane-tomlinson Went through all emails and verified links have UTM params we talked about, this includes the alternative link. Mind giving this a r?

mailer.js
```diff
@@ -390,12 +434,17 @@ module.exports = function (log) {
   Mailer.prototype.postVerifyEmail = function (message) {
     log.trace({ op: 'mailer.postVerifyEmail', email: message.email, uid: message.uid })
 
+    var templateName = 'postVerifyEmail'
     // special utm params, just for this email
     // details at github.com/mozilla/fxa-auth-mailer/issues/110
-    var postVerifyUtmParams = '?utm_source=email&utm_medium=email&utm_campaign=fx-account-verified'
```
vbudhram: @rfk Looks like previously we used `fx` as the prefix, should we reuse this to keep consistency or rename to `fxa`? Currently, you have the ability to override the params when building links so we could keep them different, but would prefer them the same.

rfk: Consistency is a hobby of mine, so I agree they should all be the same.  Let's keep them all `fx-` so we don't mess up any existing metrics.

mailer.js
```diff
@@ -11,6 +12,24 @@ var DEFAULT_LOCALE = 'en'
...
+    'postVerifyEmail': 'account-verified',
+    'newDeviceLoginEmail': 'new-device-signin',
+    'recoveryEmail': 'forgot-password',
+    'resetEmail': 'forgot-password',
+    'suspiciousLocationEmail': 'suspicious-location',
+    'verifyEmail': 'welcome',
+    'verifyLoginEmail': 'new-signin',
+    'verificationReminderFirstEmail': 'hello-again',
+    'verificationReminderSecondEmail': 'still-there',
+    'verificationReminderEmail': 'hello-again'
```
shane-tomlinson: What's the difference between `verificationReminerEmail` and `verificationReminderFirstEmail`?

shane-tomlinson: Should each entry contain distinct campaign values?

mailer.js
```diff
@@ -11,6 +12,24 @@ var DEFAULT_LOCALE = 'en'
...
+    'passwordChangedEmail': 'password-changed-success',
+    'passwordResetEmail': 'password-reset',
+    'postVerifyEmail': 'account-verified',
+    'newDeviceLoginEmail': 'new-device-signin',
+    'recoveryEmail': 'forgot-password',
+    'resetEmail': 'forgot-password',
+    'suspiciousLocationEmail': 'suspicious-location',
+    'verifyEmail': 'welcome',
+    'verifyLoginEmail': 'new-signin',
+    'verificationReminderFirstEmail': 'hello-again',
```
shane-tomlinson: For first and second, should we use the same prefix and append a `1`/`2` or `first`/`second` suffix?

mailer.js
```diff
@@ -11,6 +12,24 @@ var DEFAULT_LOCALE = 'en'
...
 
 module.exports = function (log) {
+  // Email template to UTM campaign map
+  var templateNameToCampaignMap = {
+    'passwordResetRequiredEmail': 'password-reset-required',
+    'passwordChangedEmail': 'password-changed-success',
+    'passwordResetEmail': 'password-reset',
+    'postVerifyEmail': 'account-verified',
+    'newDeviceLoginEmail': 'new-device-signin',
+    'recoveryEmail': 'forgot-password',
```
shane-tomlinson: What's the difference between `recoveryEmail` and `resetEmail`?

mailer.js
```diff
@@ -11,6 +12,24 @@ var DEFAULT_LOCALE = 'en'
 var DEFAULT_TIMEZONE = 'Etc/UTC'
 
 module.exports = function (log) {
+  // Email template to UTM campaign map
+  var templateNameToCampaignMap = {
+    'passwordResetRequiredEmail': 'password-reset-required',
+    'passwordChangedEmail': 'password-changed-success',
```
shane-tomlinson: How come `password-changed-success` has the `-success` suffix, but `password-reset` does not?

mailer.js
```diff
@@ -70,6 +89,7 @@ module.exports = function (log) {
     this.iosUrl = config.iosUrl
     this.supportUrl = config.supportUrl
     this.signInUrl = config.signInUrl
+    this.privacyUrl = config.privacyUrl
```
shane-tomlinson: While you are digging in here, it seems a good time to sort these in alphabetical order so they are easy to find while scanning.

mailer.js
```diff
@@ -198,9 +219,11 @@ module.exports = function (log) {
     if (message.redirectTo) { query.redirectTo = message.redirectTo }
     if (message.resume) { query.resume = message.resume }
 
-    var link = this.verificationUrl + '?' + qs.stringify(query)
+    var link = this._generateUTMLink(this.verificationUrl, query, templateName, 'activate')
```
shane-tomlinson: It seems to me the pattern of generating 2/3/4/5 links in sequence could be collapsed into a helper function that generates the primary link, an alternative link, a one click link, and perhaps the support links. The function would return an object with several properties, `link`, `alternative, `oneClick`, `support`, etc.

mailer.js
```diff
@@ -250,25 +278,30 @@ module.exports = function (log) {
...
+        supportLinkAttributes: this._supportLinkAttributes(templateName),
+        supportUrl: this.createSupportLink(templateName),
         timestamp: this._constructLocalTimeString(message.timeZone, message.acceptLanguage)
       },
       uid: message.uid
     })
   }
 
   Mailer.prototype.recoveryEmail = function (message) {
+    // TODO: There seems to be some discrepancy in email template names, some places
```
shane-tomlinson: Ahha, that's the answer to [this question above](https://github.com/mozilla/fxa-auth-mailer/pull/190/files/8ec8d82a8b88556c4379fa038e8c06db65a4a953#r73328105). Time to unify nomenclature?

mailer.js
```diff
@@ -485,34 +540,72 @@ module.exports = function (log) {
...
+
+    if (link) {
+      // Check if url contains any query params, if so, append new params and return
+      var parsedLink = url.parse(link)
+      if (parsedLink.query) {
+        return parsedLink.href + '&' + qs.stringify(query)
+      }
+    }
+
+    return link + '?' + qs.stringify(query)
```
shane-tomlinson: Can `link` ever be `null` or `undefined`? If so, what does this line return? I see the check for `if (link)` above.

mailer.js
```diff
@@ -485,34 +540,72 @@ module.exports = function (log) {
...
+  }
+
+  Mailer.prototype.createPasswordResetLink = function (email, templateName) {
+    var query = { email: email, reset_password_confirm: false }
```
shane-tomlinson: What does `reset_password_confirm` signify here? [Documentation about the significance of each of the possible query parameters](https://github.com/mozilla/fxa-content-server/blob/master/docs/query-params.md) would be helpful, even if only at the top of this file.

mailer.js
```diff
@@ -485,34 +540,72 @@ module.exports = function (log) {
...
-    return this.initiatePasswordResetUrl + '?' + qs.stringify(queryParams)
+    if (content) {
+      query['utm_content'] = utmPrefix + content
+    }
+
+    if (link) {
+      // Check if url contains any query params, if so, append new params and return
+      var parsedLink = url.parse(link)
+      if (parsedLink.query) {
+        return parsedLink.href + '&' + qs.stringify(query)
```
shane-tomlinson: Are there any edge cases doing this? If the original query params and the new query params both contain the same key, the result will have duplicate keys. Are you better off merging the two objects and then doing qs.stringify on the merged object?

mailer.js
```diff
@@ -11,6 +12,24 @@ var DEFAULT_LOCALE = 'en'
...
+    'recoveryEmail': 'forgot-password',
+    'resetEmail': 'forgot-password',
+    'suspiciousLocationEmail': 'suspicious-location',
+    'verifyEmail': 'welcome',
+    'verifyLoginEmail': 'new-signin',
+    'verificationReminderFirstEmail': 'hello-again',
+    'verificationReminderSecondEmail': 'still-there',
+    'verificationReminderEmail': 'hello-again'
+  }
+  var utmPrefix = 'fxa-'
```
shane-tomlinson: Similar to DEFAULT_TIMEZONE above, this is a named constant that renames the same, could you rename it to `UTM_PREFIX` for clarity? I had to search to find where this was declared when I saw it used below.

templates/new_device_login.txt
```diff
@@ -5,8 +5,8 @@
 {{#if location}}{{ location }}{{/if}}
 {{#if timestamp}}{{ timestamp }}{{/if}}
 
-{{t "This is an automated email; if you didn't add a new device to your Firefox Account, you should change your password immediately at %(resetLink)s."}}
-{{t "For more information, please visit %(supportUrl)s"}}
+{{{t "This is an automated email; if you didn't add a new device to your Firefox Account, you should change your password immediately at %(resetLink)s."}}}
```
shane-tomlinson: Wow, I'm suprised we didn't notice this before. The link would have been HTML escaped.

mailer.js
vbudhram: This is a little confusing, [verificationReminerEmail](https://github.com/mozilla/fxa-auth-mailer/blob/master/mailer.js#L451) is actually the mailer function that gets called, and `verificationReminderFirstEmail`, `verificationReminderSecondEmail` are the templates names which are determined by the `type` param. This is the only mailer function that can send different email types depending on a param.
 
Would it make more sense to put `verificationReminerEmail` testing in its own section, to clear up some of the confusion? I agree that each email template have its own distinct campaign.

@vladikoff Thoughts?

mailer.js
vbudhram: I like first/second

mailer.js
vbudhram: There is no difference, looks like this was a case of differences in mailer function and template names. Renaming `reset.html` to `recovery.html` clears up discrepancy. Will refactor template name.

mailer.js
vbudhram: Thinking more about this, I was trying to cover a case where you could just get back the params without a link. I don't really think it has much value, removing check.

mailer.js
vbudhram: Looks like it [shows/hides](https://github.com/mozilla/fxa-content-server/blob/17e27c10c2c40a474309ca376757882244d9164a/app/tests/spec/views/complete_reset_password.js#L209) the sync warning and sign-in buttons on the password reset page. I'll update docs.

mailer.js
vbudhram: I think we are better off merging objects with passed params taking priority.

vbudhram:
@shane-tomlinson made the updates you recommended, this what you had in mind? I found doing a side to side diff the easiest way to view changes.

Only outstanding question is about how to handle `verificationReminderEmails` [comment](https://github.com/mozilla/fxa-auth-mailer/pull/190#discussion_r73356677)

mailer.js
```diff
@@ -278,109 +300,120 @@ module.exports = function (log) {
     if (message.redirectTo) { query.redirectTo = message.redirectTo }
     if (message.resume) { query.resume = message.resume }
 
-    var link = this.passwordResetUrl + '?' + qs.stringify(query)
+    var links = this._generateLinks(this.passwordResetUrl, message.email, query, templateName, 'reset-password')
```
shane-tomlinson: This is so much cleaner! The only other refinement I can think of here, is `content` always going to be the same for a given `templateName`? If so, would it make sense to have a `templateName` => `content` lookup table?

shane-tomlinson:
Other than the last comment, this looks good! Can you use the write-to-disk.js script to generate all the emails and ensure they still do what they are supposed to do?

vbudhram:
@shane-tomlinson Tested all emails using the write-to-disk.js script, and they work as expected. I also published the emails to https://www.emailonacid.com/app/cp/dashboard for your viewing pleasure.

However, I couldn't get a fully deployed version working correctly because of this [issue](https://github.com/mozilla/fxa-dev/issues/262), but it is partial deployed here, https://vbudhramutm3.dev.lcip.org/signup.

templates/password_reset_required.txt
```diff
@@ -11,4 +11,4 @@
 {{{t "The Firefox Accounts team" }}}
```
shane-tomlinson: Does the link on line 8 need UTM params?

templates/new_device_login.txt
```diff
@@ -5,8 +5,8 @@
 {{#if location}}{{ location }}{{/if}}
 {{#if timestamp}}{{ timestamp }}{{/if}}
 
-{{t "This is an automated email; if you didn't add a new device to your Firefox Account, you should change your password immediately at %(resetLink)s."}}
-{{t "For more information, please visit %(supportUrl)s"}}
+{{{t "This is an automated email; if you didn't add a new device to your Firefox Account, you should change your password immediately at %(passwordChangeLink)s."}}}
+{{{t "For more information, please visit %(supportUrl)s"}}}
```
shane-tomlinson: Can you add a space above this line? It looks squashed when viewing the text:

```

This is an automated email; if you didn't add a new device to your Firefox Account, you should change your password immediately at http://127.0.0.1:3030/settings/change_password?email=testuser%40testuser.com&utm_source=email&utm_medium=email&utm_campaign=fx-new-device-signin&utm_content=fx-change-password.
For more information, please visit https://support.mozilla.org/kb/im-having-problems-with-my-firefox-account?utm_source=email&utm_medium=email&utm_campaign=fx-new-device-signin&utm_content=fx-support
```

templates/new_device_login.txt
```diff
@@ -5,8 +5,8 @@
 {{#if location}}{{ location }}{{/if}}
 {{#if timestamp}}{{ timestamp }}{{/if}}
 
-{{t "This is an automated email; if you didn't add a new device to your Firefox Account, you should change your password immediately at %(resetLink)s."}}
-{{t "For more information, please visit %(supportUrl)s"}}
+{{{t "This is an automated email; if you didn't add a new device to your Firefox Account, you should change your password immediately at %(passwordChangeLink)s."}}}
```
shane-tomlinson: I know the trailing period was there before, when copying the link out of the email users won't know whether that period should be there or not. If the user copies the link w/ the period, the utm_content value will be `fx-change-password.`

For reference, the utm_content is currently:

```
&utm_content=fx-change-password.
```

templates/password_reset.txt
```diff
@@ -2,7 +2,7 @@
 {{t "Your Firefox Account password has been successfully reset." }}
 
 {{{t "This is an automated email; if you didn’t reset the password to your Firefox Account, you should reset it immediately at %(resetLink)s." }}}
-{{t "For more information, please visit %(supportUrl)s" }}
+{{{t "For more information, please visit %(supportUrl)s" }}}
```
shane-tomlinson: Can you add a space above this line? With the new lengthy reset link, this looks squashed.

templates/password_reset.txt
```diff
@@ -2,7 +2,7 @@
 {{t "Your Firefox Account password has been successfully reset." }}
 
 {{{t "This is an automated email; if you didn’t reset the password to your Firefox Account, you should reset it immediately at %(resetLink)s." }}}
```
shane-tomlinson: Can you remove the trailing `.` here too?

templates/recovery.txt
```diff
@@ -4,7 +4,7 @@
 {{t "Remember password? Sign in."}} {{{ signInUrl }}}
 {{t "Reset password"}} {{{link}}}
```
shane-tomlinson: Can you add a space above this line? With the long links this looks squashed too:

```
Remember password? Sign in. http://127.0.0.1:3030/signin?email=testuser%40testuser.com&utm_source=email&utm_medium=email&utm_campaign=fx-forgot-password&utm_content=fx-remember-password
Reset password http://127.0.0.1:3030/v1/complete_reset_password?token=47b22cd271963448cf36da95cccfcfb342b5693d66f58aa635f9a95579431002&code=ae35999f861ffc81d594034eb4560af8&email=testuser%40testuser.com&service=sync&redirectTo=https%3A%2F%2Fredirect.com%2F&resume=eyJjYW1wYWlnbiI6bnVsbCwiZW50cnlwb2ludCI6bnVsbCwiZmxvd0lkIjoiM2Q1ODZiNzY4Mzc2NGJhOWFiNzhkMzMxMTdjZDU4Y2RmYjk3Mzk5MWU5NTk0NjgxODBlMDUyMmY2MThhNmEyMSIsInJlc2V0UGFzc3dvcmRDb25maXJtIjp0cnVlLCJ1bmlxdWVVc2VySWQiOiI1ODNkOGFlYS00NzU3LTRiZTQtYWJlNC0wZWQ2NWZhY2Y2YWQiLCJ1dG1DYW1wYWlnbiI6bnVsbCwidXRtQ29udGVudCI6bnVsbCwidXRtTWVkaXVtIjpudWxsLCJ1dG1Tb3VyY2UiOm51bGwsInV0bVRlcm0iOm51bGx9&utm_source=email&utm_medium=email&utm_campaign=fx-forgot-password&utm_content=fx-reset-password
```

templates/password_changed.txt
```diff
@@ -2,7 +2,7 @@
 {{t "Your Firefox Account password has been successfully changed." }}
 
 {{{t "This is an automated email; if you didn’t change the password to your Firefox Account, you should reset it immediately at %(resetLink)s." }}}
```
shane-tomlinson: Can you remove the trailing `.` here too?

shane-tomlinson:
@vbudhram - this PR is in really good shape. I had a look at all of the emails locally and left comments, primarily about the text emails containing trailing `.` after the links and spacing.

Once those are fixed, merge away!

templates/password_reset_required.txt
vbudhram: Done!

vbudhram:
@shane-tomlinson Thanks for review!

PR #198, opened 2016-08-10
feat(mailer): Add signin authorization and support.

### LINKED ISSUE #196: Add signin-authorization email
Ref https://docs.google.com/document/d/1qbzfA21T0HoHpgxXlpO26qcCQDx2v0PUbcoHW9XtxQQ

Could use some @ryanfeeley foo.

Initial mockup from @rfk:

screen shot 2016-08-08 at 12 00 07

shane-tomlinson:
fixes #196

### DIFF
--- a/config.js
+++ b/config.js
@@ -218,5 +218,6 @@ conf.set('mail.passwordResetUrl', contentServerUrl + '/v1/complete_reset_passwor
 conf.set('mail.initiatePasswordResetUrl', contentServerUrl + '/reset_password')
 conf.set('mail.initiatePasswordChangeUrl', contentServerUrl + '/settings/change_password')
 conf.set('mail.verifyLoginUrl', contentServerUrl + '/complete_signin')
+conf.set('mail.reportSignInUrl', contentServerUrl + '/reject_unblock_code')
 
 module.exports = conf
--- a/mailer.js
+++ b/mailer.js
@@ -296,7 +296,7 @@ module.exports = function (log) {
         'X-Unblock-Code': message.unblockCode,
         'X-Report-SignIn-Link': links.reportSignInLink
       },
-      subject: gettext('Firefox Account Authorization Code'),
+      subject: gettext('Firefox Account authorization code'),
       template: templateName,
       templateValues: {
         device: this._formatUserAgentInfo(message),
@@ -501,7 +501,7 @@ module.exports = function (log) {
       headers: {
         'X-Link': links.link
       },
-      subject: gettext('Firefox Account Verified'),
+      subject: gettext('Firefox Account verified'),
       template: templateName,
       templateValues: {
         alternativeLink: links.alternativeLink,
--- a/mailer.js
+++ b/mailer.js
@@ -23,6 +23,7 @@ module.exports = function (log) {
     'postVerifyEmail': 'account-verified',
     'recoveryEmail': 'forgot-password',
     'suspiciousLocationEmail': 'suspicious-location',
+    'unblockCode': 'new-unblock',
     'verifyEmail': 'welcome',
     'verifyLoginEmail': 'new-signin',
     'verificationReminderFirstEmail': 'hello-again-first',
@@ -40,6 +41,7 @@ module.exports = function (log) {
     'postVerifyEmail': 'connect-device',
     'recoveryEmail': 'reset-password',
     'suspiciousLocationEmail': 'password-reset',
+    'unblockCode': 'unblock-code',
     'verificationReminderFirstEmail': 'activate',
     'verificationReminderSecondEmail': 'activate',
     'verificationReminderEmail': 'activate',
@@ -102,6 +104,7 @@ module.exports = function (log) {
     this.passwordManagerInfoUrl = config.passwordManagerInfoUrl
     this.passwordResetUrl = config.passwordResetUrl
     this.privacyUrl = config.privacyUrl
+    this.reportSignInUrl = config.reportSignInUrl
     this.sender = config.sender
     this.supportUrl = config.supportUrl
     this.syncUrl = config.syncUrl
@@ -268,6 +271,43 @@ module.exports = function (log) {
         privacyUrl: links.privacyUrl,
         supportUrl: links.supportUrl,
         supportLinkAttributes: links.supportLinkAttributes
+      },
+      uid: message.uid
+    })
+  }
+
+  Mailer.prototype.unblockCodeEmail = function (message) {
+    log.trace({ op: 'mailer.unblockCodeEmail', email: message.email, uid: message.uid })
+
+    var templateName = 'unblockCodeEmail'
+    var query = {
+      unblockCode: message.unblockCode,
+      email: message.email,
+      uid: message.uid
+    }
+
+    var links = this._generateLinks(null, message.email, query, templateName)
+
+    return this.send({
+      acceptLanguage: message.acceptLanguage,
+      email: message.email,
+      headers: {
+        'X-Uid': message.uid,
+        'X-Unblock-Code': message.unblockCode,
+        'X-Report-SignIn-Link': links.reportSignInLink
+      },
+      subject: gettext('Firefox Account Authorization Code'),
+      template: templateName,
+      templateValues: {
+        device: this._formatUserAgentInfo(message),
+        email: message.email,
+        ip: message.ip,
+        location: this._constructLocationString(message),
+        privacyUrl: links.privacyUrl,
+        reportSignInLink: links.reportSignInLink,
+        reportSignInLinkAttributes: links.reportSignInLinkAttributes,
+        timestamp: this._constructLocalTimeString(message.timeZone, message.acceptLanguage),
+        unblockCode: message.unblockCode
       },
       uid: message.uid
     })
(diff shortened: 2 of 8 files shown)


mailer.js
```diff
@@ -263,6 +255,52 @@ module.exports = function (log) {
...
+    log.trace({ op: 'mailer.authoirzeLoginEmail', email: message.email, uid: message.uid })
+
+    var templateName = 'authorizeLoginEmail'
+    var query = {
+      code: message.code,
+      email: message.email,
+      uid: message.uid
+    }
+
+    if (message.service) { query.service = message.service }
```
shane-tomlinson: These three parameters are no longer needed.

mailer.js
```diff
@@ -263,6 +255,52 @@ module.exports = function (log) {
     })
   }
 
+
+  Mailer.prototype.authorizeLoginEmail = function (message) {
```
shane-tomlinson: Would `unblockCodeEmail` be a better name for this?

seanmonstar: Perhaps. I wouldn't say objectively, but naming is hard.

mailer.js
```diff
@@ -263,6 +255,52 @@ module.exports = function (log) {
...
+    log.trace({ op: 'mailer.authoirzeLoginEmail', email: message.email, uid: message.uid })
+
+    var templateName = 'authorizeLoginEmail'
+    var query = {
+      code: message.code,
```
shane-tomlinson: @seanmonstar - do these three parameters work for you? Would `code` be better as `unblockCode`?

mailer.js
```diff
@@ -263,6 +255,52 @@ module.exports = function (log) {
...
+    var links = this._generateLinks(this.authorizeLoginUrl, message.email, query, templateName)
+
+    return this.send({
+      acceptLanguage: message.acceptLanguage,
+      email: message.email,
+      headers: {
+        'X-Link': links.link,
+        'X-Service-ID': message.service,
+        'X-Uid': message.uid,
+        'X-Verify-Code': message.code
```
shane-tomlinson: `Verify` -> `Unblock`

mailer.js
seanmonstar: Should I delete them from the design doc for `POST /account/login/send_unblock_code` as well?

mailer.js
seanmonstar: These work. I'm find with `code`. Or maybe it feels cleaner all being the same. I don't lean either way, other than the easier way, which may be "just leave it".

mailer.js
shane-tomlinson: Yeah, please. Since there is no longer a link to click on, service, redirectTo and resume are no longer needed.

mailer.js
```diff
@@ -263,6 +255,52 @@ module.exports = function (log) {
...
+    if (message.redirectTo) { query.redirectTo = message.redirectTo }
+    if (message.resume) { query.resume = message.resume }
+
+    var links = this._generateLinks(this.authorizeLoginUrl, message.email, query, templateName)
+
+    return this.send({
+      acceptLanguage: message.acceptLanguage,
+      email: message.email,
+      headers: {
+        'X-Link': links.link,
```
shane-tomlinson: X-Link and X-Service-Id can both be ditched.

config.js
```diff
@@ -218,5 +218,6 @@ conf.set('mail.passwordResetUrl', contentServerUrl + '/v1/complete_reset_passwor
 conf.set('mail.initiatePasswordResetUrl', contentServerUrl + '/reset_password')
 conf.set('mail.initiatePasswordChangeUrl', contentServerUrl + '/settings/change_password')
 conf.set('mail.verifyLoginUrl', contentServerUrl + '/complete_signin')
+conf.set('mail.reportLinkUrl', contentServerUrl + '/report')
```
shane-tomlinson: `reportLinkUrl` or `rejectUnblockCodeUrl`? Could we make it a generic report link? Convert this to `rejectUnblockCodeUrl` for now.

shane-tomlinson:
@seanmonstar - could you give this an r and we can start ticking off some items?

config.js
```diff
@@ -218,5 +218,6 @@ conf.set('mail.passwordResetUrl', contentServerUrl + '/v1/complete_reset_passwor
 conf.set('mail.initiatePasswordResetUrl', contentServerUrl + '/reset_password')
 conf.set('mail.initiatePasswordChangeUrl', contentServerUrl + '/settings/change_password')
 conf.set('mail.verifyLoginUrl', contentServerUrl + '/complete_signin')
+conf.set('mail.reportSignInUrl', contentServerUrl + '/reject_unblock_code')
```
shane-tomlinson: I wasn't sure whether to call this `reportSignInUrl` to be generic and allow for future expansion, or just simply `rejectUnblockCode`.

seanmonstar: I like the generic name.

mailer.js
```diff
@@ -628,6 +673,18 @@ module.exports = function (log) {
     return this._generateUTMLink(this.initiatePasswordChangeUrl, query, templateName, 'change-password')
   }
 
+  Mailer.prototype.createReportSignInLink = function (email, templateName, data) {
```
seanmonstar: Seems `email` isn't needed in this method?

mailer.js
```diff
@@ -272,6 +273,43 @@ module.exports = function (log) {
...
+
+    return this.send({
+      acceptLanguage: message.acceptLanguage,
+      email: message.email,
+      headers: {
+        'X-Uid': message.uid,
+        'X-Unblock-Code': message.unblockCode,
+        'X-Report-SignIn-Link': links.reportSignInLink
+      },
+      subject: gettext('Firefox Account authorization code'),
```
seanmonstar: I assume you've already worked on this with @ryanfeeley, but it stuck out at me: no title case for a subject?

shane-tomlinson: To tell you the truth, I didn't think about it. @ryanfeeley - title case?

mailer.js
shane-tomlinson: Good catch, updating.

mailer.js
```diff
@@ -272,6 +273,43 @@ module.exports = function (log) {
     })
   }
 
+  Mailer.prototype.unblockCodeEmail = function (message) {
+    log.trace({ op: 'mailer.unblockCodeEmail', email: message.email, uid: message.uid })
+
+    var templateName = 'unblockCodeEmail'
```
vbudhram: You can probably add `unblockCodeEmai` to the UTM [params map](https://github.com/mozilla/fxa-auth-mailer/pull/198/files#diff-3427c5e3d78d8d6550178ad46b9e0680R18). Not sure of a good name though? Maybe `new-unblock`?

shane-tomlinson: This email doesn't actually have a primary link, only a report link. Needed?

shane-tomlinson: Cool, will add it then.

shane-tomlinson:
This is a strange one to land. The auth-server needs this PR for its tests to pass, yet if we land this PR before the auth-server, when starting up the auth server, it'll complain about the missing `reportSignInUrl` config option.

@seanmonstar - ideas to work around this? Could we merge a small PR to the auth server that sets a value for `smtp.reportSignInUrl`, then land this, then land the rest of the auth-server changes?

mailer.js
shane-tomlinson: @ryanfeeley Asked Me to Title Case It, So I did.

seanmonstar:
That does seem odd. Doesn't the `npm-shrinkwrap.json` lock to a commit, and so it shouldn't update to a newer mailer until we tell it to?

seanmonstar:
This is what I meant: https://github.com/mozilla/fxa-auth-server/blob/master/npm-shrinkwrap.json#L893

shane-tomlinson:
@seanmonstar - you are totally correct, I forgot about that. I'm the only one that symlinks to my dev fxa-auth-mailer and has to deal with the complaint when switching branches.

shane-tomlinson:
@ryanfeeley - here is a screen shot of the email

screen shot 2016-09-29 at 22 04 01

It seems to me the code should stand out a bit more, perhaps be larger or a different color. There also seems like the code has an awful lot of whitespace around it. Feedback?

@seanmonstar - from a coding P.O.V., mind reviewing?

ryanfeeley:
@shane-tomlinson Looks good. Make sure that if any extra white space is copied when the numbers are selected that our field can handle it. For example, we could accept more than 8 characters in the input, but delete any white space.

vladikoff:
The code still looks kinda small to me, but 
image

shane-tomlinson:
> Looks good. Make sure that if any extra white space is copied when the numbers are selected that our field can handle it. For example, we could accept more than 8 characters in the input, but delete any white space.


Thanks @ryanfeeley - handled on the content server as of this morning.

scripts/write-to-disk.js
```diff
@@ -101,7 +101,7 @@ function sendMail(mailer, messageToSend) {
     uaBrowserVersion: '51',
     uaOS: 'Mac OSX',
     uaOSVersion: '10.11',
-    unblockCode: '12AZ5P33',
+    unblockCode: '1ILO0Z5P',
```
vladikoff: 🍔

PR #204, opened 2016-09-06
fix(links): Remove non-functional "sign in" link.

### LINKED ISSUE #203: "Remember password? Sign in" link may not work?
Our password-reset email contains a "Remember password? Sign in" link:

screen shot 2016-08-31 at 11 30 11

With a url like the following:

```
https://accounts.firefox.com/signin?email=rfkelly%40mozilla.com&utm_source=email&utm_medium=email&utm_campaign=fx-forgot-password&utm_content=fx-remember-password
```

It's not obvious to me whether this link will do anything useful.  Following it will let you sign in to Firefox Accounts on the web, but since it doesn't contain e.g. `context=fx_desktop_v3` or similar, I don't think it will enable you to actually sign in to sync.  And if we can't cause the user to be properly signed in, perhaps we shouldn't have this link at all.

@shane-tomlinson @ryanfeeley thoughts?

rfk:
Fixes #203; @vbudhram r?

### DIFF
--- a/config.js
+++ b/config.js
@@ -179,11 +179,6 @@ var conf = convict({
       format: String,
       default: 'https://app.adjust.com/2uo1qc?campaign=fxa-conf-email&adgroup=ios&creative=button&fallback=https%3A%2F%2Fitunes.apple.com%2Fapp%2Fapple-store%2Fid989804926%3Fpt%3D373246%26ct%3Dadjust_tracker%26mt%3D8'
     },
-    signInUrl: {
-      doc: 'Deprecated. uses contentServer.url',
-      format: String,
-      default: undefined
-    },
     supportUrl: {
       doc: 'url to Mozilla Support product page',
       format: String,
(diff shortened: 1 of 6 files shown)


rfk:
actually @shane-tomlinson may be a better r? as he has existing context on this issue :-)

config.js
```diff
@@ -179,11 +179,6 @@ var conf = convict({
       format: String,
       default: 'https://app.adjust.com/2uo1qc?campaign=fxa-conf-email&adgroup=ios&creative=button&fallback=https%3A%2F%2Fitunes.apple.com%2Fapp%2Fapple-store%2Fid989804926%3Fpt%3D373246%26ct%3Dadjust_tracker%26mt%3D8'
     },
-    signInUrl: {
```
shane-tomlinson: I believe removing this here will also require a corresponding [change to the auth-server](https://github.com/mozilla/fxa-auth-server/blob/master/config/index.js#L227) and [here too](https://github.com/mozilla/fxa-auth-server/blob/master/config/index.js#L495), and possibly puppet-config.

shane-tomlinson:
Other than the comment about also updating the auth-server and possibly puppet-config, lgtm!

Thanks @rfk. Back to you.

config.js
rfk: Nice catch, thanks.  I can't find any references to `signInUrl` in puppet-config so I think we're safe there.

PR #206, opened 2016-09-07
changes to make this ready for production

jrgm:
As a start, I have these changes to get this ready for production. There will probably be other changes later, but this is a first set.

1. dump the full config at startup.
2. errors should be logged at error level.
3. Use language list from fxa-shared.

r? - @vladikoff


vladikoff:
@jrgm looks good!

PR #208, opened 2016-09-07
fix(send): improve error logging

### LINKED ISSUE #207: No error reported/handled on failure to send SMTP message
There's a bit of a serious bug here: https://github.com/mozilla/fxa-auth-mailer/blob/master/lib/verification-reminders.js#L31-L37

If that attempt to send mail fails (e.g., MTA refuses the connection, or refuses auth creds, or a misconfiguration tries to send the email to something that is not listening, etc., etc.), the error is not handled and nothing but success is reported in the logs (and in the case of verification reminders, the reminder record is removed). I'll leave it to you to how best to properly handle the error. But since this code is somewhat patterned on previous code, I'm curious if there are other paths which don't handle an error from Mailer.prototype.send.

It might be useful also to change https://github.com/mozilla/fxa-auth-mailer/blob/master/mailer.js#L212 from `log.trace` to `log.error`.

vladikoff:
Fixes #207

@jrgm r?

### DIFF
--- a/mailer.js
+++ b/mailer.js
@@ -218,17 +218,16 @@ module.exports = function (log) {
               id: status && status.messageId
             }
           )
-          d.reject(err)
-        } else {
-          log.info(
-            {
-              op: 'mailer.send.1',
-              status: status && status.message,
-              id: status && status.messageId
-            }
-          )
-          d.resolve(status)
+          return d.reject(err)
         }
+        log.info(
+          {
+            op: 'mailer.send.1',
+            status: status && status.message,
+            id: status && status.messageId
+          }
+        )
+        return d.resolve(status)
       }
     )
     return d.promise
--- a/mailer.js
+++ b/mailer.js
@@ -209,15 +209,26 @@ module.exports = function (log) {
     this.mailer.sendMail(
       emailConfig,
       function (err, status) {
-        log.trace(
-          {
-            op: 'mailer.send.1',
-            err: err && err.message,
-            status: status && status.message,
-            id: status && status.messageId
-          }
-        )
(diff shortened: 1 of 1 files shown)


mailer.js
```diff
@@ -209,15 +209,26 @@ module.exports = function (log) {
...
+            {
+              op: 'mailer.send.1',
+              err: err && err.message,
+              status: status && status.message,
+              id: status && status.messageId
+            }
+          )
+          d.reject(err)
+        } else {
+          log.trace(
```
vladikoff: @jrgm is `info` better than `trace` here or naaah?

jrgm: Actually, I'd prefer to be verbose about the success of the actual send (so `info`). (Note: `trace` is effectively `/dev/null` in production).

vladikoff: sounds good 👍

mailer.js
```diff
@@ -209,15 +209,26 @@ module.exports = function (log) {
...
+        if (err) {
+          log.error(
+            {
+              op: 'mailer.send.1',
+              err: err && err.message,
+              status: status && status.message,
+              id: status && status.messageId
+            }
+          )
+          d.reject(err)
```
jrgm: I know the code paths are either-or, but I kinda prefer the pattern:
```
  if (err) {
    log.error(...)
    return d.reject(err)
  }
  log.info(...)
  return d.resolve(status)
```
My 2c.

jrgm:
With my comment, which you can decline to take, r+

vladikoff:
Thanks @jrgm !

PR #213, opened 2016-09-20
feat(headers): Add email template name to headers sent

### LINKED ISSUE #209: Add template name as email header
To better help us understand which emails are getting bounced and why, we should add the email template name as a header. This information can then be logged via the [bouncer](https://github.com/mozilla/fxa-auth-server/blob/master/lib/bounces.js#L70). 

Reference: http://sesblog.amazon.com/post/Tx2KD4OAUNKLD51/Amazon-SES-Now-Includes-Original-Message-Headers-in-Notifications

This supports https://github.com/mozilla/fxa-auth-server/issues/1425

vbudhram:
Fixes #209

### DIFF
--- a/mailer.js
+++ b/mailer.js
@@ -195,7 +195,8 @@ module.exports = function (log) {
       text: localized.text,
       html: localized.html,
       headers: extend({
-        'Content-Language': localized.language
+        'Content-Language': localized.language,
+        'template': message.template
       }, message.headers)
     }
 
--- a/mailer.js
+++ b/mailer.js
@@ -196,7 +196,7 @@ module.exports = function (log) {
       html: localized.html,
       headers: extend({
         'Content-Language': localized.language,
-        'template': message.template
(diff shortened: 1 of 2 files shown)


vbudhram:
@vladikoff This what you were thinking? r?

mailer.js
```diff
@@ -195,7 +195,8 @@ module.exports = function (log) {
       text: localized.text,
       html: localized.html,
       headers: extend({
-        'Content-Language': localized.language
+        'Content-Language': localized.language,
+        'template': message.template
```
vladikoff: might need an `X-` 

Examples:
```
X-Forwarded-For

X-Powered-By: PHP/5.2.6-2ubuntu4.2
X-Pingback: http://www.example.org/xmlrpc.php 
```

vbudhram:
@vladikoff Updated PR, was their anything else you think we should add in?

vladikoff:
👍

PR #215, opened 2016-09-29
feat(geo): add state info into emails

### LINKED ISSUE #201: Add missing state information on geo profile emails
Something I noticed in our geo profiled emails, is that it doesn't include state information for US.

screen shot 2016-08-19 at 1 34 25 pm

I would expect this to be read as `Orlando, Florida, United States (estimated)` instead of `Orlando, United States (estimated)`.

@ryanfeeley Thoughts?

vladikoff:
Fixes #201

@vbudhram r?

### DIFF
--- a/mailer.js
+++ b/mailer.js
@@ -154,8 +154,12 @@ module.exports = function (log) {
     var location = message.location
     // construct the location string from the location object
     if (location) {
-      if (location.city) {
+      if (location.city && location.stateCode) {
+        return translator.format(translator.gettext('%(city)s, %(stateCode)s, %(country)s (estimated)'), location)
+      } else if (location.city) {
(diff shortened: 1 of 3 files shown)


mailer.js
```diff
@@ -154,8 +154,12 @@ module.exports = function (log) {
     var location = message.location
     // construct the location string from the location object
     if (location) {
-      if (location.city) {
+      if (location.city && location.stateCode) {
+        return translator.format(translator.gettext('%(city)s, %(stateCode)s, %(country)s (estimated)'), location)
```
vbudhram: Ack looks like you found a gap in our coverage. Mind adding test cases in for 

* city and country
* state and country
* only country

PR #218, opened 2016-09-29
fix(coverage): Add script to enabled code coverage for local tests

vbudhram:
Noticed test coverage wasn't being reported for this. Uses the common coverage script.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -44,6 +44,6 @@ install:
 script:
   - grunt templates
   - if [ $DB == "mysql" ]; then ./scripts/start-travis-auth-db-mysql.sh; fi
-  - npm test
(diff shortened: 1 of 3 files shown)


vbudhram:
@mozilla/fxa-devs r?

vladikoff:
@vbudhram you probably want to enable coveralls for this as well, like: 

https://github.com/mozilla/fxa-auth-server/blob/master/.travis.yml#L53

vladikoff:
@vbudhram I enabled it here: https://coveralls.io/github/mozilla/fxa-auth-mailer

vbudhram:
@vladikoff Code coverage added, is there anything else I should add?

PR #219, opened 2016-09-29
fix(mailer): Swap the order of IP and Location.

### LINKED ISSUE #216: Reverse line order of IP and location in our emails
I got some feedback from the UX team about the upcoming geo-location information.

Currently, in this order, we show:

* IP
* Location (if available)

Preferably, in this order, we should show:

* Location (if available)
* IP

Easy peasy @shane-tomlinson squeezy?

shane-tomlinson:
Extract the common stuff to partials so it's easier to change from
now on.

fixes #216 

@vbudhram - mind an r?

note to @shane-tomlinson - update the template in #198 to use the partial.

### DIFF
--- a/templates/new_device_login.txt
+++ b/templates/new_device_login.txt
@@ -1,8 +1,8 @@
 {{t "New sign-in to Firefox" }}
 
 {{ device }}
-{{#if ip}}{{t "IP address: %(ip)s" }}{{/if}}
 {{#if location}}{{ location }}{{/if}}
+{{#if ip}}{{t "IP address: %(ip)s" }}{{/if}}
 {{#if timestamp}}{{ timestamp }}{{/if}}
 
 {{{t "This is an automated email; if you didn’t add a new device to your Firefox Account, you should change your password immediately at %(passwordChangeLink)s"}}}
--- a/templates/verify_login.html
+++ b/templates/verify_login.html
@@ -21,10 +21,12 @@
   <td valign="top">
     <h1 style="font-family: sans-serif; font-weight: normal; margin: 0 0 24px 0; text-align: center;">{{t "New sign-in to Firefox"}}</h1>
     <p class="primary" style="font-family: sans-serif; font-size: 14px; font-weight: normal; margin: 0 0 24px 0; text-align: center;">{{t "For added security, please confirm this sign-in to begin syncing with this device:"}}<br/><br/>
-    {{{ device }}}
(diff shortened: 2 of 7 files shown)


templates/new_device_login.txt
```diff
@@ -1,8 +1,8 @@
 {{t "New sign-in to Firefox" }}
 
 {{ device }}
-{{#if ip}}{{t "IP address: %(ip)s" }}{{/if}}
 {{#if location}}{{ location }}{{/if}}
+{{#if ip}}{{t "IP address: %(ip)s" }}{{/if}}
```
seanmonstar: This cannot use the partial?

seanmonstar: Oh right of course, this is the text version, partial has HTML

templates/verify_login.html
```diff
@@ -21,10 +21,12 @@
...
     <h1 style="font-family: sans-serif; font-weight: normal; margin: 0 0 24px 0; text-align: center;">{{t "New sign-in to Firefox"}}</h1>
     <p class="primary" style="font-family: sans-serif; font-size: 14px; font-weight: normal; margin: 0 0 24px 0; text-align: center;">{{t "For added security, please confirm this sign-in to begin syncing with this device:"}}<br/><br/>
-    {{{ device }}}
-    {{#if ip }}<span style="display:block;margin:2px 0;">{{t "IP address: %(ip)s" }}</span>{{/if}}
-    {{#if location }}<span style="display:block;margin:2px 0;">{{ location }}</span>{{/if}}
-    {{#if timestamp }}<span style="display:block;margin:2px 0;">{{ timestamp }}</span>{{/if}}
+    <span style="display:block;margin:2px 0;">{{{ device }}}</span>
+{{#if location }}<span style="display:block;margin:2px 0;">{{ location }}</span>{{/if}}
+{{#if ip }}<span style="display:block;margin:2px 0;">{{t "IP address: %(ip)s" }}</span>{{/if}}
+{{#if timestamp }}<span style="display:block;margin:2px 0;">{{ timestamp }}</span>{{/if}}
```
seanmonstar: It seems this could use the partial?

shane-tomlinson: All the .html files in `templates` are generated using `grunt templates` from the nunjucks templates in `partials`. He. We have a double template system, nunjucks templates are used to create handlebars templates that are rendered by the mailer.

PR #220, opened 2016-09-29
fix(mailer): Fix the title size in the password reset related templates.

### LINKED ISSUE #217: Forgot your password title size mismatch
I noticed that the title of the "Forgot your password?" email does not match the size of our other emails. Correctable?

shane-tomlinson:
Change both "password_reset" and "recovery" to use `h1` instead of `h3`

fixes #217 

screen shot 2016-09-29 at 21 54 49

screen shot 2016-09-29 at 21 54 53

@ryanfeeley - r?

### DIFF
--- a/partials/password_reset.html
+++ b/partials/password_reset.html
@@ -4,7 +4,7 @@
 <!--Header Area-->
 <tr style="page-break-before: always">
     <td valign="top">
-        <h3 style="font-family: sans-serif; font-weight: normal; margin: 0 0 24px 0; text-align: center;">{{t "Your password has been reset"}}</h3>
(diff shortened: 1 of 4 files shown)


TDA:
@shane-tomlinson Doesnt this look too big? Can we reach a middle ground with h2?

ryanfeeley:
@TDA I would love to simplify the code in our HTML emails if possible. I have proposed some styles above, but we'd really need to test in all clients via emailonacid to see if it works  https://docs.google.com/document/d/1I7TSNa7mvUFipX48OhGSRbuVybIqSb5QdnCmivSjl-c/edit

TDA:
@ryanfeeley I dont have permissions to see it :(

TDA:
@ryanfeeley I took a look at those, they look really good. When you say `simplify the code`, what exactly do you have in mind? streamlining the font sizes/colors? or both? or am I completely off about what you are thinking?

PR #229, opened 2016-10-18
feat(mailer): Remove the `Alternatively` link from all emails.

### LINKED ISSUE #228: Remove the "Alternatively" links in the email.
Request from @ryanfeeley.

shane-tomlinson:
It's big. It detracts from the main link. It's so big that it's
impossible to place other content below it. @ryanfeeley wants it out!

fixes #228

@vbudhram - r?

### DIFF
--- a/mailer.js
+++ b/mailer.js
@@ -268,7 +268,6 @@ module.exports = function (log) {
       subject: gettext('Verify your Firefox Account'),
       template: templateName,
       templateValues: {
-        alternativeLink: links.alternativeLink,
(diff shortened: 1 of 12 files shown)


vbudhram:
@shane-tomlinson LGTM! Look's like removing lines lowered our overall coverage to below 94%. I think this is ok to merge and file as follow up?

screen shot 2016-10-18 at 11 23 23 am

vladikoff:
Wow!

PR #236, opened 2016-11-13
fix(config): disable polling if pollTime === 0

jrgm:
r? - @vladikoff or @jbuck 

The database queries are extremely unfriendly to mysql replication. I have disabled `verificationReminders.rate` in the auth-server, but I'd like a graceful way to disable this polling entirely today, and we can revisit what other changes are needed on Monday.

### DIFF
--- a/config.js
+++ b/config.js
@@ -112,7 +112,7 @@ var conf = convict({
       default: 20
     },
     pollTime: {
-      doc: 'Poll duration in milliseconds.',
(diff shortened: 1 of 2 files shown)


jrgm:
> The database queries are extremely unfriendly to mysql replication.

FYI: We have a read replica for DR purposes, if you wondering "what replication?".

vladikoff:
lgtm!!

PR #241, opened 2016-12-16
fix(metrics): Add flowId and flowBeginTime to headers if avalible

vbudhram:
This PR adds support for forwarding the `flowId` and `flowBeginTime` values in the email headers. These will be used to track email bounce events that correspond to flows. This has been added to all our emails except `verificationReminderEmail`.

Part of https://github.com/mozilla/fxa-auth-server/issues/1425

### DIFF
--- a/mailer.js
+++ b/mailer.js
@@ -268,15 +268,22 @@ module.exports = function (log) {
 
     var links = this._generateLinks(this.verificationUrl, message.email, query, templateName)
 
+    var headers = {
+      'X-Link': links.link,
+      'X-Service-ID': message.service,
+      'X-Uid': message.uid,
+      'X-Verify-Code': message.code
+    }
+
+    if (message.flowBeginTime && message.flowId) {
+      headers['X-Flow-Id'] = message.flowId
+      headers['X-Flow-Begin-Time'] = message.flowBeginTime
+    }
+
     return this.send({
       acceptLanguage: message.acceptLanguage,
       email: message.email,
-      headers: {
-        'X-Link': links.link,
-        'X-Service-ID': message.service,
-        'X-Uid': message.uid,
-        'X-Verify-Code': message.code
-      },
+      headers: headers,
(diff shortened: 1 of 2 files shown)


seanmonstar:
This looks great to me.

Did we really need the `flowBeginTime`? I'd expect that to have been part of the initial flow creation, and then we wouldn't need it again. All time lengths can be computed by subtracting the original stored begin time in the DB.

mailer.js
```diff
@@ -268,15 +268,22 @@ module.exports = function (log) {
 
     var links = this._generateLinks(this.verificationUrl, message.email, query, templateName)
 
+    var headers = {
```
seanmonstar: Nit if you want, I guess we prefer `let` now over `var` wherever possible. If I didn't have the question about `flowBeginTime`, I would have merged regardless, so however you like.

rfk:
> Did we really need the flowBeginTime?

Yes, although one could argue that it's mostly for historical reasons at this point.  You need to know the flowBeginTime to validate the signature in the flowId, and we use it to calculate and log the time between start of flow and each particular event.

You're right that this second usage could be done in post-processing based on data in the DB, but that wasn't obvious when we were designing these events initially.  If it becomes a PITA to keep passing this timestamp around then we could try to refactor ourselves towards doing it in the DB.

mailer.js
vbudhram: Updated this to use `let` but the l10 extraction complains that it is a reserved keyword. @vladikoff is this something we should update to exclude `let` or should we hold off on using it in repo for now?

vbudhram:
@rfk Made https://github.com/mozilla/fxa-auth-server/issues/1594 to help track instances we might need to refactor`flowBeginTime` and help build a case for doing it.

vladikoff:
from mtg: don't use `let` for now.

vbudhram:
@seanmonstar Mind an official r?

Issue #189, opened 2016-07-26
Add location information to verify_email, reset password and password changed templates


shane-tomlinson:
For example, password reset is sometimes used to compromise accounts. Perhaps we should show location information to these emails too.

cc @ryanfeeley - if you don't think it'd provide any value, feel free to close.

ryanfeeley:
@shane-tomlinson Current emails. Are we sending suspicious sign-in notices yet? https://docs.google.com/document/d/1I7TSNa7mvUFipX48OhGSRbuVybIqSb5QdnCmivSjl-c/edit#

rfk:
> Are we sending suspicious sign-in notices yet? 

No, and per [1] I have no reason to believe we ever will.  The user gets more timely notice about login attempts from the "new login" and/or "confirm signin" emails.

[1] https://github.com/mozilla/fxa-auth-mailer/issues/183#issuecomment-234924763

vladikoff:
@ryanfeeley which emails should we add this to?

ryanfeeley:
On reset request, you could add the location information after replacing the line of copy above it to be: Click the button within the next hour to set a new password for your Firefox Account. The request came from the following device:

ryanfeeley:
Proposal for verification email: <img width="399" alt="screenshot 2016-12-16 14 28 58" src="https://cloud.githubusercontent.com/assets/68213/21276949/d2cbb996-c3a1-11e6-8748-220939e4f190.png">

vladikoff:
From mtg:

Add the location info into the following:

* verification email with text tweaks
* "reset your password"

ryanfeeley:
from password change notification…

CURRENT

* Your Firefox Account password has been successfully changed.

PROPOSED

* Your Firefox Account password has been successfully changed from this device: (include info)

ryanfeeley:
kill date too

ryanfeeley:
<img width="642" alt="image" src="https://cloud.githubusercontent.com/assets/68213/21278496/809be224-c3a9-11e6-9318-78256523b0b1.png">
https://docs.google.com/document/d/1I7TSNa7mvUFipX48OhGSRbuVybIqSb5QdnCmivSjl-c/edit#

seanmonstar:
> kill date too

While the date doesn't seem useful many times since the timestamp for when the email received is probably pretty close, it seems like it is useful for when delivery was delayed a while.

vbudhram:
[...]

PR #244, opened 2016-12-23
fix(strings): add comma to ready, set, sync

vladikoff:


vbudhram:
Ref: https://github.com/mozilla/fxa-auth-mailer/issues/189

PR #247, opened 2017-01-12
set up "SES Event Publishing" metrics

jrgm:
This (along with changes to come for fxa-auth-server, and config changes in AWS), will provide more fine-grained details on SES usage.

r? - @vladikoff, @vbudhram

### DIFF
--- a/mailer.js
+++ b/mailer.js
@@ -12,6 +12,9 @@ var DEFAULT_LOCALE = 'en'
 var DEFAULT_TIMEZONE = 'Etc/UTC'
 var UTM_PREFIX = 'fx-'
 
+var X_SES_CONFIGURATION_SET = 'X-SES-CONFIGURATION-SET'
+var X_SES_MESSAGE_TAGS = 'X-SES-MESSAGE-TAGS'
+
 module.exports = function (log) {
   // Email template to UTM campaign map, each of these should be unique and
   // map to exactly one email template.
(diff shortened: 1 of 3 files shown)


mailer.js
```diff
@@ -276,6 +291,10 @@ module.exports = function (log) {
       'X-Verify-Code': message.code
     }
 
+    if (this.sesConfigurationSet) {
+      headers[X_SES_MESSAGE_TAGS] = sesMessageTagsHeaderValue(templateName)
```
vbudhram: Since this is being sent for all emails, we could move it into `Mailer.prototype.send`. Something like 

```javascript
    if (this.sesConfigurationSet) {
      // Note on SES Event Publishing: The X-SES-CONFIGURATION-SET and
      // X-SES-MESSAGE-TAGS email headers will be stripped by SES from the
      // actual outgoing email messages.
      emailConfig.headers[X_SES_CONFIGURATION_SET] = this.sesConfigurationSet
      emailConfig.headers[X_SES_MESSAGE_TAGS] = sesMessageTagsHeaderValue(message.template)
}
```

jrgm: So, `.template` is not set on `message`, at least for `verifyEmail`. So I can't consolidate things that way.

vbudhram: @jrgm Sounds good, for some reason thought that it was always set.

vbudhram:
@jrgm One more nit is that a test case could be added similar to [header checks](https://github.com/mozilla/fxa-auth-mailer/blob/master/test/local/mailer_tests.js#L123).

jrgm:
> @jrgm One more nit is that a test case could be added similar to header checks.

I'm confused by this. Isn't that the test I already added?

vbudhram:
> I'm confused by this. Isn't that the test I already added?

@jrgm Oops...you are correct, I didn't see the changes in that file. Merge at will!

PR #251, opened 2017-01-27
fix(dependencies): update bluebird, nodemailer, convict, moment-timezone

seanmonstar:
Code changes were due to updates in bluebird, it no longer uses `cancellable`, but it's easy enough to just have a local variable track that.

### DIFF
--- a/lib/db_connect.js
+++ b/lib/db_connect.js
@@ -19,6 +19,7 @@ var dbBackend = config.get('db').backend
  */
 module.exports = function () {
   function dbConnect() {
+    var cancelled = false
     function dbConnectPoll() {
       return DB.connect(config.get(dbBackend))
         .then(
@@ -28,7 +29,14 @@ module.exports = function () {
           function (err) {
             if (err && err.message && err.message.indexOf('ECONNREFUSED') > -1) {
               log.warn('db', {message: 'Failed to connect to database, retrying...'})
-              return P.delay(dbConnectionRetry).then(dbConnectPoll)
+              if (!cancelled) {
+                return P.delay(dbConnectionRetry)
+                  .then(function () {
+                    if (!cancelled) {
+                      return dbConnectPoll()
+                    }
+                  })
+              }
             } else {
               log.error('db', {err: err})
             }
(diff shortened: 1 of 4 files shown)


lib/db_connect.js
```diff
@@ -28,7 +31,14 @@ module.exports = function () {
           function (err) {
             if (err && err.message && err.message.indexOf('ECONNREFUSED') > -1) {
               log.warn('db', {message: 'Failed to connect to database, retrying...'})
-              return P.delay(dbConnectionRetry).then(dbConnectPoll)
+              if (!cancelled) {
+                return P.delay(dbConnectionRetry)
+                  .then(() => {
```
vladikoff: @seanmonstar i think l10n extract cannot parse this, so you either need to change this to use `function` or update to  l10n modules. I suggest just removing the `() => {`  and we can update the l10n modules later....

seanmonstar: I was so confused at that error. Oh well, fixed.

vladikoff:
with the new mailer i get:

```
fxa-auth-server.ERROR: geodb.1 {"op":"geodb.1","err":"Unable to fetch data"}
fxa-auth-server.INFO: mailer.send {"email":"xeq@adas.com","op":"mailer.send","template":"verifyEmail"}
fxa-auth-server.ERROR: mailer.send.1 {"op":"mailer.send.1","err":"Unsupported configuration, downgrade Nodemailer to v0.7.1 to use it"}
fxa-auth-server.ERROR: mailer.sendVerifyCode.1 {"op":"mailer.sendVerifyCode.1","err":"Error: Unsupported configuration, downgrade Nodemailer to v0.7.1 to use it"}
fxa-auth-server.ERROR: account.create.confirm.error {"op":"account.create.confirm.error","uid":"79a762090f87454880ad4720ac2deb3e","err":"Error: Unsupported configuration, downgrade Nodemailer to v0.7.1 to use it","tokenVerificationId":{"type":"Buffer","data":[235,229,100,8,115,213,232,246,6,210,6,151,56,25,251,57]}}
```

vladikoff:
@seanmonstar getting several `Unsupported configuration, downgrade Nodemailer to v0.7.1 to use it"`

vladikoff:
@seanmonstar we can probably merge this and then start a PR on the auth server to support it?

seanmonstar:
I'll merge this later today unless someone sees something bizarre.

vladikoff:
Yeah I tested this with a local maildev and it was fine 👍 

Thanks @seanmonstar !

PR #254, opened 2017-02-01
feat(mailer): add support for sending SMS messages

philbooth:
Related to mozilla/fxa-auth-server#1628.

- [x] Define the API.
- [x] Include SMS messages in the auth mailer templates for l10n.
- [x] Implement phone number validation.
- [x] Unit tests.
- [x] Fix `Invalid Text Message` error from Nexmo.
- [x] Test out sending actual messages to the UK, the USA and Canada.
- [ ] Export the sms interface for direct calling.
- [ ] Include the proper install link.
- [ ] Include the finalised message text.
- [ ] Include the finalised sender ids.

/cc @shane-tomlinson @vbudhram


lib/sms-sender-ids.json
```diff
@@ -0,0 +1,5 @@
+[
+  [ "CA", "16474909977" ],
```
shane-tomlinson: What're these?

lib/sms.js
```diff
@@ -0,0 +1,89 @@
...
+
+module.exports = (log, translator, templates, config) => {
+  const nexmo = new Nexmo({
+    apiKey: config.apiKey,
+    apiSecret: config.apiSecret
+  })
+  const sendSms = P.promisify(nexmo.message.sendSms, { context: nexmo.message })
+
+  // TODO: Sort out the real link
+  const LINK = 'https://moz.la/vwxyz'
```
shane-tomlinson: Ya, we need a real link. @davismtl - what adjust link should we point people at? Should we run it through the moz.la shortener?

lib/sms.js
```diff
@@ -0,0 +1,89 @@
...
+          }
+        })
+    }
+  }
+
+  function getSenderId (phoneNumber) {
+    const numberUtil = libpn.PhoneNumberUtil.getInstance()
+    const parsedNumber = numberUtil.parse(phoneNumber)
+
+    if (! numberUtil.isValidNumber(parsedNumber)) {
```
shane-tomlinson: I'm assuming this will also be done on the auth server.

lib/sms.js
```diff
@@ -0,0 +1,89 @@
...
+          const resultCount = result.messages && result.messages.length
+          if (resultCount !== 1) {
+            log.error('sms.send', { err: new Error('Unexpected result count'), result })
+          }
+
+          result = result.messages[0]
+          const status = result.status
+
+          if (status !== '0') {
+            fail(`Delivery failed: ${status} ${result['error-text']}`, 500)
```
shane-tomlinson: This error will be returned to the auth-server I'm guessing. What does the auth-server do with it? Does the auth-server convert errors to the standardized format where each error gets its own errno?

lib/sms.js
```diff
@@ -0,0 +1,89 @@
...
+const Nexmo = require('nexmo')
+const P = require('bluebird')
+const SENDER_IDS = new Map(require('./sms-sender-ids.json'))
+
+module.exports = (log, translator, templates, config) => {
+  const nexmo = new Nexmo({
+    apiKey: config.apiKey,
+    apiSecret: config.apiSecret
+  })
+  const sendSms = P.promisify(nexmo.message.sendSms, { context: nexmo.message })
```
shane-tomlinson: This may be pre-mature, any value in passing in the `sendSms` function so that we can swap out providers if need be, e.g., if we decide to use Basket or Amazon SNS instead of Nexmo? Basket may duplicate a lot of the functionality here, I'm not sure how it works.

templates/sms.1.html
```diff
@@ -0,0 +1 @@
+sms.1.txt
```
shane-tomlinson: What's this file for? Necessary to load the templates?

lib/sms-sender-ids.json
philbooth: Virtual numbers that I'm using for testing.

As per [the documentation](https://docs.nexmo.com/messaging/sms-api/building-global-apps#senderID), legal values for the sender id depend on the region we're sending to. For the UK we can use an alphanumeric string up to 11 characters long, hence `Firefox`. For the USA it has to be a phone number, so I just plugged in the virtual phone number I'm using for USA testing, for now. For Canada, we should be able to use the short code that was already associated with out Nexmo account. However, in practice I found this got rejected by the network provider, so I've switched to the virtual number I'm using for testing there too.

For all of this, note the last item in the issue check-list: `Include the finalised sender ids`. It's one of the things we need to pin down before this can be merged.

lib/sms.js
philbooth: I don't see the point in doing it now. It's a possibility I'm aware of and that we spoke about, but writing code to handle multiple SMS APIs feels like wasted cycles at this stage. I don't foresee it being difficult to do, if/when we need it. None of the implementation/provider details leak outside of this module.

lib/sms.js
philbooth: The auth server work isn't finished yet. When it is, it will return an appropriate error.

lib/sms.js
philbooth: Why?

templates/sms.1.html
philbooth: The general template code assumes HTML versions of every template. Obviously that makes no sense for SMS, so I made it a symlink, which is what this is. But it doesn't need to be a symlink, it could be an empty file, or anything really.

Or I could add some conditional logic to the template code that ignores HTML files for `sms.*` templates, but that felt a bit like unnecessary complexity when I thought about it before.

lib/sms.js
philbooth: Or did you mean that you assume *some validation* will be done on the auth server, rather than explicitly this validation?

I do have rudimentary validation at the API level, yes. `/^\+[1-9]\d{1,14}$/` is the regex, if you're interested.

shane-tomlinson: > Or did you mean that you assume some validation will be done on the auth server, 

Yes.

lib/sms.js
davismtl: Hey, I recommend looking at this document: https://docs.google.com/spreadsheets/d/151_fWwEPxmPN_AbHCp-sqAkshDCLaABRrVaowxoFS10/edit#gid=0

This is our Adjust master tracker for owned channels. You should both have edit access now.

There are some rows with FxA in there. You will see the nomenclature of the URLs.

I recommend shortner too since it would be too long in SMS.

davismtl: And here are the deeplinking parameters too for the next phases: https://docs.adjust.com/en/deeplinking/

PR #255, opened 2017-02-03
fix(reminders): update reminder logic to support latest db

vladikoff:
Connects to https://github.com/mozilla/fxa-auth-db-mysql/issues/196

Ref: https://github.com/mozilla/fxa-auth-db-mysql/pull/200


vladikoff:
Not needed

PR #259, opened 2017-02-08
feat(mailer): add support for sending SMS messages

philbooth:
Related to mozilla/fxa-auth-server#1628. Replaces #254.

This is broadly the same as the previous PR. The main difference is that I didn't realise the auth server invokes the mailer in-process before, so some redundant server nonsense has been removed. For the same reason, the interface for the package needed changing to return the `sms` module. I opted to return an object `{ email, sms }` but am open to other suggestions if anyone finds that offensive.

With the elimination of my imaginary server boundary, it made sense to pull the config and phone number validation up to the auth server level so those are both removed. And with the config gone, the `send-sms` script has moved over there too.

All that means the code here should be a bit simpler than it was. I think it's ready for review now but note that **we don't want to merge this until the auth server changes are also ready to be merged**. The aforementioned interface change requires equivalent changes in the auth server. 

You can see those auth server changes in the [phil/issue-1628 branch](https://github.com/mozilla/fxa-auth-server/compare/master...phil/issue-1628). I just need to sort out the correct sender id for USA/Canada and wire in the flow events before it, too, is ready to be reviewed.

@mozilla/fxa-devs r?

### DIFF
--- a/lib/sms.js
+++ b/lib/sms.js
@@ -1,0 +1,98 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+'use strict'
+
+var Nexmo = require('nexmo')
+var P = require('bluebird')
+
+var TEMPLATE_NAMES = new Map([
+  [ 1, 'installFirefox' ]
+])
+
+module.exports = function (log, translator, templates, smsConfig) {
+  var nexmo = new Nexmo({
+    apiKey: smsConfig.apiKey,
+    apiSecret: smsConfig.apiSecret
+  })
+  var sendSms = P.promisify(nexmo.message.sendSms, { context: nexmo.message })
+
+  return {
+    send: function (phoneNumber, senderId, messageId, acceptLanguage) {
+      log.trace({
+        op: 'sms.send',
+        senderId: senderId,
+        messageId: messageId,
+        acceptLanguage: acceptLanguage
+      })
+
+      return P.resolve()
+        .then(function () {
+          var message = getMessage(messageId, acceptLanguage)
+
+          return sendSms(senderId, phoneNumber, message.trim())
+        })
+        .then(function (result) {
+          var resultCount = result.messages && result.messages.length
+          if (resultCount !== 1) {
+            // I don't expect this condition to be entered, certainly I haven't
+            // seen it in testing. But because I'm making an assumption about
+            // the result format, I want to log an error if my assumption proves
+            // to be wrong in production.
+            log.error({ op: 'sms.send', err: new Error('Unexpected result count'), resultCount: resultCount })
+          }
+
+          result = result.messages[0]
+          var status = result.status
+
+          // https://docs.nexmo.com/messaging/sms-api/api-reference#status-codes
+          if (status === '0') {
+            log.info({
+              op: 'sms.send.success',
+              senderId: senderId,
+              messageId: messageId,
+              acceptLanguage: acceptLanguage
+            })
+          } else {
+            var reason = result['error-text']
+            fail('Message rejected', {
+              status: 500,
+              reason: reason,
+              reasonCode: status
+            })
+          }
+        })
+    }
+  }
+
+  function getMessage (messageId, acceptLanguage) {
+    var templateName = TEMPLATE_NAMES.get(messageId)
+    var template = templates['sms.' + templateName]
+
+    if (! template) {
+      fail('Invalid message id', { status: 400 })
+    }
+
+    return template({
+      link: smsConfig[templateName + 'Link'],
+      translator: translator(acceptLanguage)
+    }).text
+  }
+
+  // If/when fxa-auth-mailer is moved into the auth server repo,
+  // calls to this function can be replaced with the auth server's
+  // AppError methods directly.
+  function fail (message, properties) {
+    log.error({ op: 'sms.send', err: message })
+
+    var error = new Error(message)
+    if (properties) {
+      Object.keys(properties).forEach(function (key) {
+        error[key] = properties[key]
+      })
+    }
+
+    throw error
+  }
+}
--- a/templates/index.js
+++ b/templates/index.js
@@ -13,12 +13,16 @@ handlebars.registerHelper(
   }
 )
 
-function camelize(str) {
+function generateTemplateName (str) {
+  if (/^sms\.[A-Za-z]+/.test(str)) {
+    return str
+  }
+
   return str.replace(/_(.)/g,
     function(match, c) {
       return c.toUpperCase()
     }
-  )
+  ) + 'Email'
 }
 
 function loadTemplates(name) {
@@ -33,7 +37,7 @@ function loadTemplates(name) {
       var renderText = handlebars.compile(text)
       var renderHtml = handlebars.compile(html)
       return {
-        name: camelize(name) + 'Email',
+        name: generateTemplateName(name),
         fn: function (values) {
           return {
             text: renderText(values),
@@ -54,6 +58,7 @@ module.exports = function () {
       'password_reset_required',
       'post_verify',
       'recovery',
+      'sms.installFirefox',
       'unblock_code',
       'verification_reminder_first',
       'verification_reminder_second',
--- a/index.js
+++ b/index.js
@@ -4,6 +4,7 @@
 
 var P = require('bluebird')
 var createMailer = require('./mailer')
+var createSms = require('./lib/sms')
 
 module.exports = function (log, config, sender) {
   var Mailer = createMailer(log)
@@ -15,7 +16,10 @@ module.exports = function (log, config, sender) {
   )
   .spread(
     function (translator, templates) {
-      return new Mailer(translator, templates, config.mail, sender)
+      return {
+        email: new Mailer(translator, templates, config.mail, sender),
+        sms: createSms(log, translator, templates, config.sms)
+      }
     }
   )
 }
--- a/npm-shrinkwrap.json
+++ b/npm-shrinkwrap.json
@@ -1476,7 +1476,7 @@
                 },
                 "minimist": {
                   "version": "1.2.0",
-                  "from": "minimist@>=1.1.3 <2.0.0",
+                  "from": "minimist@1.2.0",
                   "resolved": "https://registry.npmjs.org/minimist/-/minimist-1.2.0.tgz"
                 },
                 "normalize-package-data": {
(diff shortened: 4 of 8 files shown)


lib/sms.js
```diff
@@ -0,0 +1,84 @@
...
+var P = require('bluebird')
+
+module.exports = function (log, translator, templates, config) {
+  var nexmo = new Nexmo({
+    apiKey: config.apiKey,
+    apiSecret: config.apiSecret
+  })
+  var sendSms = P.promisify(nexmo.message.sendSms, { context: nexmo.message })
+
+  var LINK = 'http://mzl.la/1HOd4ec'
```
vladikoff: should we make this part of config? (just in case we need to tweak this, change the page and redeploy)

Also, can we use HTTPS for this? :)

lib/sms.js
```diff
@@ -0,0 +1,84 @@
...
+
+  return {
+    send: function (phoneNumber, senderId, messageId, acceptLanguage) {
+      log.info({
+        op: 'sms.send',
+        phoneNumber: phoneNumber,
```
vladikoff: not sure about PII here, we either need to update the filter in Kibana or not log the phone number.
if we cannot log the full number, maybe just the area code?

lib/sms.js
```diff
@@ -0,0 +1,84 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+'use strict'
+
+var Nexmo = require('nexmo')
+var P = require('bluebird')
+
+module.exports = function (log, translator, templates, config) {
```
vladikoff: nit: `config` -> `smsConfig`, so it is clear that it is not the full server config. nit though!

lib/sms.js
```diff
@@ -0,0 +1,84 @@
...
+      return P.resolve()
+        .then(function () {
+          var message = getMessage(messageId, acceptLanguage)
+
+          return sendSms(senderId, phoneNumber, message.trim())
+        })
+        .then(function (result) {
+          var resultCount = result.messages && result.messages.length
+          if (resultCount !== 1) {
+            log.error({ op: 'sms.send', err: new Error('Unexpected result count'), result: result })
```
vladikoff: we might benefit from a "success!" log here with either `.info` (or trace). I think we already use `.info` for successful Push notifications

lib/sms.js
```diff
@@ -0,0 +1,84 @@
...
+
+var Nexmo = require('nexmo')
+var P = require('bluebird')
+
+module.exports = function (log, translator, templates, config) {
+  var nexmo = new Nexmo({
+    apiKey: config.apiKey,
+    apiSecret: config.apiSecret
+  })
+  var sendSms = P.promisify(nexmo.message.sendSms, { context: nexmo.message })
```
vladikoff: nit: do we need a quick doc string here about what `context: nexmo.message` is?

lib/sms.js
vladikoff: Does `result: result ` contain PII?

templates/sms.1.txt
```diff
@@ -0,0 +1 @@
+{{t "As requested, here is a link to install Firefox on your mobile device: %(link)s"}}
```
vladikoff: nit: I wonder if we should bother with submitting this for translation while we still test this EN locales.
I'm not sure what the timeline is to make this available to other locales, maybe something to discuss.

shane-tomlinson: I'm of two minds, though leaning on the side of "let's ship it to the l10n team." Even US-only, it'd be nice to have EN and ES, though, if we use short-codes, we might have to store [already translated strings with the SMS provider](https://dashboard.nexmo.com/buy-numbers/shortcodes/alerts#).

templates/index.js
```diff
@@ -54,6 +58,7 @@ module.exports = function () {
       'password_reset_required',
       'post_verify',
       'recovery',
+      'sms.1',
```
vbudhram: Not advocating for or against, but why use template name as number? Others have an explicit name that make it a little easier to know what that template does.

lib/sms.js
philbooth: I'll make the change but fwiw, I coped this pattern from `mailer.js`. It receives `config.mail` and references it as `config` locally.

templates/sms.1.txt
philbooth: > I wonder if we should bother with submitting this for translation while we still test this EN locales.

Fwiw, I disagree. What is the benefit of not localizing?

> ...if we use short-codes, we might have to store already translated strings with the SMS provider.

I created a separate issue to discuss this quetion: mozilla/fxa-auth-server#1645.

philbooth:
@mozilla/fxa-devs r?

lib/sms.js
```diff
@@ -0,0 +1,86 @@
...
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+'use strict'
+
+var Nexmo = require('nexmo')
+var P = require('bluebird')
+
+var TEMPLATE_NAMES = new Map([
+  [ 1, 'installFirefox' ]
```
vbudhram: 👍

lib/sms.js
```diff
@@ -0,0 +1,86 @@
...
+  })
+  var sendSms = P.promisify(nexmo.message.sendSms, { context: nexmo.message })
+  var firefoxAppStoreLink = smsConfig.firefoxAppStoreLink
+
+  return {
+    send: function (phoneNumber, senderId, messageId, acceptLanguage) {
+      log.info({
+        op: 'sms.send',
+        senderId: senderId,
+        messageId: messageId,
```
vbudhram: I wonder if it might be useful to do a similar logging pattern as the mailer. On send, you have a [trace](https://github.com/mozilla/fxa-auth-mailer/blob/master/mailer.js#L215) level log.

lib/sms.js
```diff
@@ -0,0 +1,86 @@
...
+        acceptLanguage: acceptLanguage
+      })
+
+      return P.resolve()
```
vladikoff: @philbooth do you need this P.resolve here if `return sendSms(senderId, phoneNumber, message.trim())` already returns a Promise?

lib/sms.js
```diff
@@ -0,0 +1,86 @@
...
+        .then(function () {
+          var message = getMessage(messageId, acceptLanguage)
+
+          return sendSms(senderId, phoneNumber, message.trim())
+        })
+        .then(function (result) {
+          var resultCount = result.messages && result.messages.length
+          if (resultCount !== 1) {
+            log.error({ op: 'sms.send', err: new Error('Unexpected result count'), resultCount: resultCount })
```
vbudhram: Maybe leave a little note about what it means to have an unexpected result count. Is it really an error? My initial thought was that two messages were sent to the number.

lib/sms.js
```diff
@@ -0,0 +1,86 @@
...
+          }
+
+          result = result.messages[0]
+          var status = result.status
+
+          if (status !== '0') {
```
vladikoff: nit: @philbooth can we add a link to the nexmo docs here or something, so it is easier to understand the result we get from it?

lib/sms.js
```diff
@@ -0,0 +1,86 @@
...
+  [ 1, 'installFirefox' ]
+])
+
+module.exports = function (log, translator, templates, smsConfig) {
+  var nexmo = new Nexmo({
+    apiKey: smsConfig.apiKey,
+    apiSecret: smsConfig.apiSecret
+  })
+  var sendSms = P.promisify(nexmo.message.sendSms, { context: nexmo.message })
+  var firefoxAppStoreLink = smsConfig.firefoxAppStoreLink
```
vladikoff: nit: the name `firefoxAppStoreLink` assumes that we will always send an appStore link, worth to rename?

lib/sms.js
```diff
@@ -0,0 +1,86 @@
...
+          var status = result.status
+
+          if (status !== '0') {
+            var reason = result['error-text']
+            fail('Message rejected', {
+              status: 500,
+              reason: reason,
+              reasonCode: status
+            })
+          }
```
vbudhram: Should we log a success here? Looks like auth-server logs a flowEvent, but maybe still useful to visualize success in Kibana.

lib/sms.js
philbooth: The mailer has a `trace` on entry to the function and an `info` immediately before sending. Having both of those here seems like unnecessary log noise to me, because there's so little going on and the error cases are all logged separately. Hence I combined the two into one and discarded the lower log level, `trace`.

Doing it the opposite way, discarding `info` in favour of `trace` seems wrong. And having them both seems noisy.

lib/sms.js
philbooth: `getMessage` throws so I'm calling it inside the promise chain to convert the error to a rejection.

I did play with the alternative approach, of making everything return promises from the off, but it makes the promise chain longer and obscures some of the logic a little bit. This way seemed neater to me.

lib/sms.js
philbooth: Do you mean rename the `var` or rename the config option or rename both? Happy to do it, whichever.

What would be a better name for you, `firefoxInstallLink` or `firefoxLink` or something else?

lib/sms.js
vladikoff: sounds good!

lib/sms.js
```diff
@@ -0,0 +1,98 @@
...
+  [ 1, 'installFirefox' ]
+])
+
+module.exports = function (log, translator, templates, smsConfig) {
+  var nexmo = new Nexmo({
+    apiKey: smsConfig.apiKey,
+    apiSecret: smsConfig.apiSecret
+  })
+  var sendSms = P.promisify(nexmo.message.sendSms, { context: nexmo.message })
+  var installFirefoxLink = smsConfig.firefoxAppStoreLink
```
vladikoff: @philbooth i think we can rename the config option as well.

`smsConfig.link` or `smsLink`, I kind of like just a simple `link`. i.e "The sms sent contains this link" which can be anything, in this case a link to the app store at the moment.

philbooth: If the name is just a generic `link`, what happens in the future if we add other message types that contain different links? In that scenario, we want the specific name so that we can distinguish between them in the code and have some kind of map from message ids to link names, like there is for message ids to template names already.

Actually, following that train of thought further, it strikes me that the link name should be the same as the template name, if they are part of a pair. So both called `installFirefox` in this case, and both keyed by the message id. What do you think?

lib/sms.js
```diff
@@ -0,0 +1,98 @@
...
+  function getMessage (messageId, acceptLanguage) {
+    var templateName = TEMPLATE_NAMES.get(messageId)
+    var template = templates['sms.' + templateName]
+
+    if (! template) {
+      fail('Invalid message id', { status: 400 })
+    }
+
+    return template({
+      link: smsConfig[templateName + 'Link'],
```
philbooth: @vladikoff, in case my last comment wasn't clear, I meant this. ^

lib/sms.js
vladikoff: > So both called installFirefox in this case, and both keyed by the message id. What do you think?

Yea! let's do that!

vladikoff:
👍

PR #263, opened 2017-02-21
fix(mail): add from header for the new nodemailer

### LINKED ISSUE #262: Sender name disappeared in train 80
I noticed that the sender name "Firefox Accounts" has seemingly disappeared in train 80. Any ideas? 
screenshot 2017-02-21 10 47 55

vladikoff:
Fixes #262

### DIFF
--- a/config.js
+++ b/config.js
@@ -138,7 +138,7 @@ var conf = convict({
     },
     sender: {
       doc: 'email address of the sender',
-      default: 'accounts@firefox.com',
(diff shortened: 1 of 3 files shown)


vladikoff:
@vbudhram r?

seanmonstar:
Looks good! We should probably add a test to the auth server that makes use of the nodemailer to verify setting this option does what we hope.

